PR #8, opened 2016-08-10
Refactor xml2csv.py to run with less RAM

cuducos:
As I said, it's way slower than the previous method, but I could convert `AnosAnteriores.xml` (2.6G) into `AnosAnteriores.csv` (291M) with my MacBook (8Gb RAM, Core M) — it took more than 4 hours though.

### DIFF
--- a/src/xml2csv.py
+++ b/src/xml2csv.py
@@ -1,24 +1,89 @@
-from bs4 import BeautifulSoup
 import csv
+import gc
+import json
+import os
 import sys
+from datetime import datetime
+from tempfile import TemporaryDirectory
+
+import numpy as np
+from bs4 import BeautifulSoup
 
 XML_FILE_PATH = sys.argv[1]
 CSV_FILE_PATH = sys.argv[2]
 
-xml_file = open(XML_FILE_PATH, 'rb').read()
-xml_soup = BeautifulSoup(xml_file, 'lxml-xml')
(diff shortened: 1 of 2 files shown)


src/xml2csv.py
```diff
@@ -1,24 +1,89 @@
...
+output('Fetching CSV headers')
+with open('data/datasets_format.html', 'rb') as file_handler:
+    parsed = BeautifulSoup(file_handler.read(), 'lxml')
+    for row in parsed.select('.tabela-2 tr'):
+        try:
+            field_names.append(row.select('td')[0].text.strip())
+        except IndexError:
+            pass
+    del(parsed)
+    gc.collect()
```
Irio: Good one, didn't know about this interface.

cuducos: I've read about it but I'm not sure about how effective this actually is. 

Even after running `gc.collect()` my memory monitor still highlights a high memory usage. In spite of that I'm still able to allocate memory for other stuff.

My guess is that it might let the OS know that some blocks could be freed if needed and the OS handles it — but it's just a guess (and I haven't read the full documentation on the `gc` module either… so shame on me).

Irio:
Took 2h36min here (including ~40min when I let my machine sleep while going out for lunch), with an i7 8 cores and 16GB of memory. 🚀

Issue #14, opened 2016-08-21
Get address and coordinates of each CNPJ from expenses


Irio:
Useful for analysis #13.

Irio:
To get the dataset, already available on Amazon S3, run the following:

```sh
$ python src/fetch_datasets.py
```

PR #19, opened 2016-08-21
New dataset with companies information

Irio:
Comes to solve our need (#14) of addresses and coordinates of all companies listed in the expenses. The initial version is the result a collaboration between me, @filipelinhares and @andrepinho.

Here's a short summary of what the current dataset has (not sure if this notebook fits well in the repository, probably not): https://gist.github.com/Irio/002a1d3842466aacd0027c6f382a2d9b
Once this pull request gets approved, I will uploaded it to Amazon S3.

Current version of the dataset contains a list of 59,985 companies; when counting just the Brazilian ones, 3,453 are missing given timeouts returned by the API used.

**Missing**

- [x] Get coordinates of addresses
- [x] Translate dataset
- [x] Upload file to Amazon S3 and add its URL to `src/fetch_datasets.py`

### DIFF
--- a/src/clean_cnpj_info_dataset.py
+++ b/src/clean_cnpj_info_dataset.py
@@ -1,6 +1,7 @@
 from itertools import chain
 import json
 import numpy as np
+import os
 import pandas as pd
 
 def decompose_main_activity(value):
@@ -24,7 +25,7 @@ def decompose_secondary_activities(value):
 
 
 
-data = pd.read_csv('data/cnpj_info.xz',
+data = pd.read_csv(os.path.join('data', 'cnpj_info.xz'),
                    dtype={'atividade_principal': np.str,
                           'atividades_secundarias': np.str,
                           'complemento': np.str,
@@ -81,4 +82,7 @@ data = pd.concat([
     data['secondary_activities'].apply(decompose_secondary_activities)],
     axis=1)
 
-data.to_csv('data/companies.xz', compression='xz', encoding='utf-8', index=False)
+data.to_csv(os.path.join('data', 'companies.xz'),
+            compression='xz',
+            encoding='utf-8',
+            index=False)
--- a/src/clean_cnpj_info_dataset.py
+++ b/src/clean_cnpj_info_dataset.py
@@ -25,7 +25,7 @@ def decompose_secondary_activities(value):
 
 
 
-data = pd.read_csv(os.path.join('data', 'cnpj_info.xz'),
+data = pd.read_csv(os.path.join('data', 'cnpj-info.xz'),
                    dtype={'atividade_principal': np.str,
                           'atividades_secundarias': np.str,
                           'complemento': np.str,
--- a/src/clean_cnpj_info_dataset.py
+++ b/src/clean_cnpj_info_dataset.py
@@ -1,0 +1,84 @@
+from itertools import chain
+import json
+import numpy as np
+import pandas as pd
+
+def decompose_main_activity(value):
+    struct = json.loads(value.replace('\'', '"'))
+    if struct:
+        return pd.Series(struct[0]). \
+            rename_axis({'code': 'main_activity_code', 'text': 'main_activity'})
+    else:
+        return pd.Series({}, index=['main_activity_code', 'main_activity'])
+
+def decompose_secondary_activities(value):
+    struct = json.loads(value.replace('\'', '"'))
+    if struct and struct[0].get('text') != 'Não informada':
+        new_attributes = [pd.Series(activity). \
+            rename_axis({'code': 'secondary_activity_%i_code' % (index + 1),
+                         'text': 'secondary_activity_%i' % (index + 1)})
+            for index, activity in enumerate(struct)]
+        return pd.concat(new_attributes)
+    else:
+        return pd.Series()
+
+
+
+data = pd.read_csv('data/cnpj_info.xz',
+                   dtype={'atividade_principal': np.str,
+                          'atividades_secundarias': np.str,
+                          'complemento': np.str,
+                          'efr': np.str,
+                          'email': np.str,
+                          'message': np.str,
+                          'motivo_situacao': np.str,
+                          'situacao_especial': np.str})
+data = data.drop_duplicates('cnpj')
+data.rename(columns={
+    'abertura': 'opening',
+    'atividade_principal': 'main_activity',
+    'atividades_secundarias': 'secondary_activities',
+    'bairro': 'neighborhood',
+    'cep': 'zip_code',
+    'complemento': 'additional_address_details',
+    'data_situacao_especial': 'special_situation_date',
+    'data_situacao': 'situation_date',
+    'efr': 'responsible_federative_entity',
+    'fantasia': 'trade_name',
+    'logradouro': 'address',
+    'motivo_situacao': 'situation_reason',
+    'municipio': 'city',
+    'natureza_juridica': 'legal_entity',
+    'nome': 'name',
+    'numero': 'number',
+    'situacao_especial': 'special_situation',
+    'situacao': 'situation',
+    'telefone': 'phone',
+    'tipo': 'type',
+    'uf': 'state',
+    'ultima_atualizacao': 'last_updated',
+}, inplace=True)
+
+categories = (
+    'legal_entity',
+    'message',
+    'responsible_federative_entity',
+    'situation_reason',
+    'situation',
+    'special_situation',
+    'status',
+    'type',
+)
+for key in categories:
+    data[key] = data[key].astype('category')
+
+data['main_activity'] = data['main_activity'].fillna('{}')
+data['secondary_activities'] = data['secondary_activities'].fillna('{}')
+
+data = pd.concat([
+    data.drop(['main_activity', 'secondary_activities'], axis=1),
+    data['main_activity'].apply(decompose_main_activity),
+    data['secondary_activities'].apply(decompose_secondary_activities)],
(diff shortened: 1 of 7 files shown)


src/clean_cnpj_info_dataset.py
```diff
@@ -0,0 +1,78 @@
+from itertools import chain
```
cuducos: `itertools.chain` 💕

src/clean_cnpj_info_dataset.py
```diff
@@ -0,0 +1,78 @@
...
+import pandas as pd
+
+def decompose_main_activity(value):
+    struct = json.loads(value.replace('\'', '"'))
+    return pd.Series([*struct]). \
+        rename_axis({'code': 'main_activity_code', 'text': 'main_activity'})
+
+def decompose_secondary_activities(value):
+    struct = json.loads(value.replace('\'', '"'))
+    if bool(struct) and struct[0]['text'] != 'Não informada':
```
cuducos: Is there a specific reason to use `bool(struct)` instead of just `struct`? We could also use `Dict.get(key)` to (eventually) avoid `KeyError`:

```python
if struct and struct[0].get('text') != 'Não informada':
    …
```

src/clean_cnpj_info_dataset.py
```diff
@@ -0,0 +1,78 @@
...
+    'cep': 'zip_code',
+    'complemento': 'additional_address_details',
+    'data_situacao_especial': 'special_situation_date',
+    'data_situacao': 'situation_date',
+    'efr': 'responsible_federative_entity',
+    'fantasia': 'trade_name',
+    'logradouro': 'address',
+    'motivo_situacao': 'situation_reason',
+    'municipio': 'city',
+    'natureza_juridica': 'legal_nature',
```
cuducos: What about `legal_entity`? Not sure about the context. Legal entity refers to the *subject* of a legal action. Legal nature refers to the sort of *object* (topic, issue, discussion, paradigm) related to law theory and philosophy. (Source: my mind — highly unreliable).

src/clean_cnpj_info_dataset.py
```diff
@@ -0,0 +1,78 @@
...
+data['legal_nature'] = data['legal_nature'].astype('category')
+data['message'] = data['message'].astype('category')
+data['responsible_federative_entity'] = \
+    data['responsible_federative_entity'].astype('category')
+data['situation_reason'] = data['situation_reason'].astype('category')
+data['situation'] = data['situation'].astype('category')
+data['special_situation'] = data['special_situation'].astype('category')
+data['status'] = data['status'].astype('category')
+data['type'] = data['type'].astype('category')
+
```
cuducos: Once more, just a matter of style (i.e. ignore it as you wish) :

```python
categories = (
    'legal_nature',
    'message',
    'responsible_federative_entity',
    'situation_reason',
    'situation',
    'special_situation',
    'status',
    'type',
)
for key in categories:
    data[key] = data[key].astype('category')
```

src/fetch_cnpj_info.py
```diff
@@ -0,0 +1,114 @@
...
+
+def remaining_cnpjs(info_dataset):
+    datasets = [read_csv(name)
+                for name in ['current-year', 'last-year', 'previous-years']]
+    dataset = pd.concat(datasets)
+    del(datasets)
+    is_cnpj = dataset['cnpj_cpf'].str.len() == 14.
+    cnpj_list = set(dataset.loc[is_cnpj, 'cnpj_cpf'])
+    already_fetched = set(info_dataset['cnpj'].str.replace(r'[./-]', ''))
+    return list(cnpj_list - already_fetched)
```
cuducos: **If** this is a huge list a generator might be useful:

```python
yield from (cnpj in cnpj_list if cnpj not in already_fetched)
```

src/fetch_cnpj_info.py
```diff
@@ -0,0 +1,114 @@
...
+        info_dataset = info_dataset.append(attributes, ignore_index=True)
+    info_dataset.to_csv(INFO_DATASET_PATH,
+                        compression='xz',
+                        encoding='utf-8',
+                        index=False)
+
+
+
+if not os.path.exists(TEMP_PATH):
+    os.makedirs(TEMP_PATH)
```
cuducos: If this is really a temporary file, we might use [`TemporaryDirectory`] (https://docs.python.org/3.5/library/tempfile.html#tempfile.TemporaryDirectory) — I can code that, pairing or not if you feel like we should use it. We can also have [an object in memory that works like a file](https://docs.python.org/3.5/library/io.html) (not sure about if this is a proper approach though… I mean, if this gets too big we'll be requiring tons of RAM).

src/fetch_cnpj_info.py
```diff
@@ -0,0 +1,114 @@
...
+
+info_dataset = load_info_dataset()
+cnpj_list = remaining_cnpjs(info_dataset)
+print('%i CNPJ\'s to be fetched' % len(cnpj_list))
+
+with futures.ThreadPoolExecutor(max_workers=10) as executor:
```
cuducos: Threads 💕

cuducos:
I added some comments in a *in line* basis. Nothing really important, but my intent was to make the code easier to understand and more optimized (when needed). Most of the comments are more stylistic than technical and accepting the suggestions or not I feel like this is good to merge already.

I'll let it up to you to merge, because I'm not sure about the 3rd item from the TODO list (Upload file to Amazon S3) — so feel free to go through the comments, calling me names for being such a dick, uploading the files and finally merging ; )

cuducos:
Rereading I found another issue — maybe more important than all the other ones.

Defining file paths like `'data/cnpj_info.xz'` might raise an error in non-Unix platforms (aka Windows). The proper way to do it is to use `os.path.join('data', 'cnpj_info.xz')` so Python uses the proper slash (`/` or `\`) according to the OS.

We might also need review our code base because this might have passed unnoticed in other PRs…

src/fetch_cnpj_info.py
Irio: I'm not sure what `yield` does here. Can you explain me?

I did it in the way above because uses `Set`'s, which's supposedly very fast for calculating differences.

cuducos: Sure thing! Sorry about that 💦

A function/method that uses `yield` instead of `return` is a *generator*. This is specially useful when it comes to memory usage. So let's go back to the diference between `list`, `set`, `tuple` (all of them) and *generators*.

When you build a `list` (or `tuple` or `set`) all its elements are calculated and are available. If you type `[0, 1, 2, 3]` this is actually stored in memory, all the four values.

However, if you type `range(4)` what is created is a *generator*, an instruction, an iterator that can be used to build a `list`, `tuple`, `set` or just used as any other iterator. It tends to use small amounts of memory than the full list, but when you iterate it, the result is the same:

```python
for n in [0, 1, 2, 3]:
    print(n)
```

Outputs the same thing as:

```python
for n in range(4):
    print(n)
```

The difference is how the iterator is created. The first example is a full list stored in memory, while the second is an instructor that calculates the [`next`](https://docs.python.org/3.5/library/functions.html#next) element **only** when (and **if** needed). Yet when the *next* value arrives, the previous one is thrown away (the previous value is not in memory anymore).

In simple examples like that the difference is not meaningful at all. But if you try to use large lists, or lists with items that are large themselves, you can feel the difference. Try these examples: the first uses a list comprehension to crate an actual `list` from a *generator*; the second example is just a generator. Try them:

```python
my_iter = [number in range(6 * 10 ** 23)]
for n in my_iter:  
    print(n)
```

```python
my_iter = range(6 * 10 ** 23)
for n in my_iter:
    print(n)
```

The first code might crash. The second one runs smoothly.

So, what I'm actually suggesting is a smart memory management: getting one CNPJ at time, just when needed. And as I commented that might be useful **if** this is a huge list.

Does it make any sense?

src/fetch_cnpj_info.py
Irio: I did it without `TemporaryDirectory` because of the use case where I run the script, companies get geolocated but for any reason the process gets finished (e.g. I need to turn the computer off, hardware or software crashes) but we don't want to lose the progress until that point. Keep in mind that geolocating 60K+ addresses takes hours, even when using multithreading. I added [this commit](https://github.com/datasciencebr/serenata-de-amor/pull/19/commits/533ddb16d4286d27fce20ccf5c286110a7d7a1b9) to delete the folder once finished.

Irio:
Squashing few commits and merging this branch now. The work hasn't completely finished, since there are 3,391 CNPJs remaining to be fetched (they keep returning 504 HTTP status). There are 57,644 geolocated companies, from a total of 60K. Some seem to be incorrectly pointed by Google Maps, but I'm happy with this first version for now.

https://nbviewer.jupyter.org/gist/Irio/22026574ebc597bdedfbdb13989ffe98

To get the dataset, already available on Amazon S3, run the following:
```sh
$ python src/fetch_datasets.py
```

src/fetch_cnpj_info.py
cuducos: I got it wrong in my first suggestion. In order to get it to work it should be:

```python
for cnpj in cnpj_list:
    if cnpj not in already_fetched:
        yield cnpj
```

Instead of:

```python
yield from (cnpj in cnpj_list if cnpj not in already_fetched)
```

src/fetch_cnpj_info.py
cuducos: Great strategy then!

PR #23, opened 2016-08-27
Refactor/optimize with lxml iterparse & StringIO

cuducos:
This new version of `xml2csv.py` is roughly 4x faster and can be ran in basically any computer (here in my computer it only uses 180Mb of RAM to run).

Basically three new things:

* This versions uses [`xml`](https://docs.python.org/3/library/xml.html)'s [`iterparse`](https://docs.python.org/3.5/library/xml.etree.elementtree.html#xml.etree.ElementTree.iterparse) (and other generators) to parse the XML without loading the full contents into memory (it means I'm not using `BeautifulSoup` to parse the XML)
* It uses the `'a'` mode (*append* mode) instead of the `'w'` mode (*write* mode) to add content to the CSV without open the whole file
* Once memory is not a problem anymore, I replaced file temporary system read & write (slow) by [`StringIO`](https://docs.python.org/3.5/library/io.html?highlight=stringio#io.StringIO) objects (that are stored in memory, i.e. they are way faster)

This PR might sound like just a simple improvement, but I do believe it is a good *strategy* to deal with large files without depending on “super computers.”

### DIFF
--- a/src/xml2csv.py
+++ b/src/xml2csv.py
@@ -1,90 +1,81 @@
-import csv
-import gc
 import json
-import os
 import sys
+from csv import DictWriter
 from datetime import datetime
-from tempfile import TemporaryDirectory
+from io import StringIO
 
-import numpy as np
 from bs4 import BeautifulSoup
+from lxml.etree import iterparse
 
 XML_FILE_PATH = sys.argv[1]
 CSV_FILE_PATH = sys.argv[2]
-
-field_names = ['ideDocumento']
-tmp_files = []
+HTML_FILE_PATH = 'data/datasets_format.html'
 
 
-def output(msg, **kwargs):
+def output(*args, **kwargs):
+    """Helper to print messages with a date/time marker"""
     now = datetime.strftime(datetime.now(), '%Y-%m-%d %H:%M:%S')
-    return print(now, msg, **kwargs)
+    return print(now, *args, **kwargs)
 
-# get extra field names
-output('Fetching CSV headers')
-with open('data/datasets_format.html', 'rb') as file_handler:
-    parsed = BeautifulSoup(file_handler.read(), 'lxml')
-    for row in parsed.select('.tabela-2 tr'):
-        try:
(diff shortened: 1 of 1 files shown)


src/xml2csv.py
```diff
@@ -1,90 +1,81 @@
...
-    with open(CSV_FILE_PATH, 'w') as csv_file:
-        writer = csv.DictWriter(csv_file, fieldnames=field_names)
+    return print(now, *args, **kwargs)
+
+
+def xml_parser(xml_path, tag='DESPESA'):
+    """
+    Generator that parses the XML yielding a StringIO object for each record
+    found. The StringIO holds the data in JSON format.
+    """
```
Irio: 👍  on adding documentation.

src/xml2csv.py
```diff
@@ -1,90 +1,81 @@
...
 import json
-import os
 import sys
+from csv import DictWriter
 from datetime import datetime
-from tempfile import TemporaryDirectory
+from io import StringIO
 
-import numpy as np
 from bs4 import BeautifulSoup
```
Irio: Just noticed: since you are not using BeautifulSoup anymore, you can remove this `import` and its requirement in the [conda_requirements.txt](https://github.com/datasciencebr/serenata-de-amor/blob/934db02ec33ddc8958ec320f0a130904f643ba3e/conda_requirements.txt) file.

Irio: My bad, you're still using it. https://github.com/datasciencebr/serenata-de-amor/pull/23/files#diff-986bb1fc5b0bcbb749606f7613c049e9R43

cuducos: Yes. I mean, to parse the HTML table BeautifulSoup still useful here and at `src/translation_table.py` — but I think we can use *here* an import and do the HTML parsing in `src/translation_table.py`. Let's merge both (as they are on separate branches) and them I look for duplicates and try to eliminate them. What do you think?

PR #24, opened 2016-08-28
Descriptive Analysis

Irio:
How the main dataset looks like? How many rows/expenses we had so far? This analysis goes through last year's (2015) dataset and tries to answer basic questions and numbers anyone who's inclined to contribute to the project may want to know.

Preview: [develop/2016-08-13-im-descriptive-analysis.ipynb](
https://github.com/datasciencebr/serenata-de-amor/blob/ffdf84298f415716e04da0128e801b94e1e53e82/develop/2016-08-13-im-descriptive-analysis.ipynb)


Irio:
Once we get this one merged, I plan to create a small library with functions like `read_csv` and `document_url` (already used in this analysis) so we don't need to have the pain of opening existing analyses to copy them from.

cuducos:
Great great job, @irio! This is awesome material.

**About negative values**

You said “not really sure what they mean”; according to the description (see PR #25) it is about unused fly tickets, but I'm also not sure about what happen. What we have is (my translation):

> Value of the expense in the receipt. If it refers to fly tickets this value can be negative, meaning that it is a credit related to another fly tickets issued but not used by the congress person (the same is valid for `net_value`)

And the original:

> O seu conteúdo é o valor de face do documento fiscal ou o valor do documento que deu causa à despesa. Quando se tratar de bilhete aéreo, esse valor poderá ser negativo, significando que o referido bilhete é um bilhete de compensação, pois compensa um outro bilhete emitido e não utilizado pelo deputado (idem para o dado vlrLiquido abaixo).

**Applicants distribution**

This distribution (`In [34]`) is very intriguing isn't it?

> Apparently, politicians from states further away from Distrito Federal expent more. We could perform an analysis on distance to the capital and the home state from the politician.

The value is different from state to state and probably the farer the state, more allowance the politician will have.

**Further ideas**

In the future we can think about an open platform to browser this data? A open web platform (not data science), but making it easy for people to *explore* this data set — even if just out of curiosity.

cuducos:
Anyway, looks good to merge from my point of view ; )

Irio:
@cuducos Great comments.

About your explanation on unused flight tickets, it means that negative values should be taken out from average calculations then. Will change the analysis before merging.

About open web platform, I feel that http://ops.net.br/ comes for solving this problem. How different would you expect it?

Issue #9, opened 2016-08-10
Translate dataset variables/description table


cuducos:
As mentioned in PR #7 (specifically [here](https://github.com/datasciencebr/serenata-de-amor/pull/7#issuecomment-238887392), [here](https://github.com/datasciencebr/serenata-de-amor/pull/7#issuecomment-238904356) and [here](https://github.com/datasciencebr/serenata-de-amor/pull/7#issuecomment-238908375))

cuducos:
Quick follow up:

I started a base for the generation of the markdown (but visualization still poor, I'm gonna to improve it) and @lrcezimbra translated the variables titles (PR #11). Therefore what is pending is:

* Improve table visualization
* Translate variable descriptions
* Maybe avoid repetition and use part of this translation file in further uses of data (cc @irio)

PR #25, opened 2016-08-28
Translation table

cuducos:
With the help of @lrcezimbra I finished the *translation table* — fixing #9.

Before merging I think we can take a look in three things:

- [x] I'd like someone to proofread the translations (they are **not** literal translation) for precision and typos
- [ ] Maybe @irio or others can use can use the dictionaries (importing this file in other scripts and using the`EN` constant; or even the `get_portuguese()` generator) to guarantee we use the same translation for the name of the variables all over the repo
- [ ] I'd like to hear from @hyena22 if this is of any help for foreigners, if things make sense, where we could (a) improve the translation and/or (b) add contextual/local data to help readers and contributors
- [ ] Should I merge these granular commits in a single one before merging…?


src/translation_table.py
```diff
@@ -0,0 +1,267 @@
...
+
+
+EN = (
+    {
+        'variable': 'congressperson_name',
+        'name': 'Congressperson Name',
+        'desc': """Name used by the congress person during his term in
+            office. Usually it is composed by two elements: a given name and a
+            family name; two given names; or two forename, except if the head
+            of the Lower House explicitly alter this rule in order to avoid
```
Irio: "Câmara dos Deputados" refer themselves as "Chamber of Deputies" in [their English version](http://www2.camara.leg.br/english). Shouldn't we prefer this other name?

Irio: From what I understand: Câmara dos Deputados/Chamber of Deputies is the lower house, while Senado/Senate is the upper.

cuducos: > From what I understand: Câmara dos Deputados/Chamber of Deputies is the lower house, while Senado/Senate is the upper.

Lower and Upper Houses are the general names. Each country have their own names and reason for it (House of Representatives / Senate in the US if I'm not wrong; House of Commons and House of Lords in UK etc.).

I tend to prefer using the general term because it is clearer for those who don't know the specificities of Brazilian political structure. But saying they refer themselves as _Chamber of Deputies_ is indeed a strong argument — let's embrace it.

Irio:
help wanted -> help given

Irio:
Thanks @lrcezimbra and @cuducos for this great work.

Issue #20, opened 2016-08-21
Collect average real estate prices per squared meter by region 


hyena22:
Retrieve estimated prices per squared meter in order to compare against declared expenses for rents by scraping from http://seriesestatisticas.ibge.gov.br/series.aspx?vcodigo=PRECO415.

janosimas:
I was looking in some other sites that can give some information on this:
http://www.agenteimovel.com.br/mercado-imobiliario/a-venda/sao-paulo,sp/
http://www.zapimoveis.com.br/quanto-vale-este-imovel-mapa-dos-precos/

luizpericolo:
@janosimas 

The first site looks really scrapable. It would be easy to build "all possible urls" to collect and save the average real estate price and no further computation would be needed since the site already returns this information. And the site returns information by neighborhood. It might not be as granular as @cuducos commented on PR #27, but it is better than having an average price by state.

The second site would be a bit harder in my opinion. Searching for some neighborhoods here in Rio and checking the response that is rendered to the map you get entries like this:

    {
  	   "AreaUtil": null 
	   "Coordenadas": {
		   "Latitude": -22.902176,
		   "Longitude": -43.280269
	   }
	   "DataUltimoAnuncio": null
	   "Dormitorios": null
	   "Endereco": null
	   "ID": 73390
	   "PrecoM2": null
	   "PrecoVenda": 271362.8263
    }

It neither returns a value for "PrecoM2" (price by square meter) nor a value for "AreaUtil" (that means something like built area). That means we don't really have information about price for square meter for any real estate and we can't calculate it either. Aside from that the body of the POST that is sent for this API relies on long and lat coordinates.

Lrcezimbra:
Looking in ZAP Imoveis, I found that search for 'estates for rent' instead of 'estates for sale', solve the problem cited by @luizpericolo, the response has a rental price (PrecoLocacao) that we can use:
```json
{
	'AreaUtil': None,
	'Coordenadas': {
		'Latitude': -30.0332536, 
		'Longitude': -51.2152337
	},
	'DataUltimoAnuncio': None,
	'Dormitorios': None,
	'Endereco': None,
	'ID': 10779682,
	'NotaLocacao': 19.2896946427,
	'NotaVenda': 0,
	'PrecoLocacao': 6500,
	'PrecoM2': None,
	'PrecoVenda': None,
	'TipoOfertaID': 2,
	'indOferta': True
}
```
[...]

PR #27, opened 2016-08-29
Collect average real estate prices

Lrcezimbra:
This PR solves issue #20?

### DIFF
--- a/src/fetch_real_estate_prices.py
+++ b/src/fetch_real_estate_prices.py
@@ -1,0 +1,10 @@
+from urllib.request import urlretrieve
+
+urls = [
+    ('region', 'http://seriesestatisticas.ibge.gov.br/exportador.aspx?arquivo=PRECO415_GR_ABS.csv&localidade=Todas'),
+    ('state', 'http://seriesestatisticas.ibge.gov.br/exportador.aspx?arquivo=PRECO415_UF_ABS.csv&localidade=Todas'),
+    ('brasil', 'http://seriesestatisticas.ibge.gov.br/exportador.aspx?arquivo=PRECO415_BR_ABS.csv',)]
+
+
+for (by, url) in urls:
+    urlretrieve(url, 'data/average_real_estate_prices_by_%s.csv' % by)


src/fetch_real_estate_prices.py
```diff
@@ -0,0 +1,10 @@
+from urllib.request import urlretrieve
+
+urls = [
+    ('region', 'http://seriesestatisticas.ibge.gov.br/exportador.aspx?arquivo=PRECO415_GR_ABS.csv&localidade=Todas'),
+    ('state', 'http://seriesestatisticas.ibge.gov.br/exportador.aspx?arquivo=PRECO415_UF_ABS.csv&localidade=Todas'),
+    ('brasil', 'http://seriesestatisticas.ibge.gov.br/exportador.aspx?arquivo=PRECO415_BR_ABS.csv',)]
+
+
+for (by, url) in urls:
+    urlretrieve(url, 'data/average_real_estate_prices_by_%s.csv' % by)
```
cuducos: This might crash in non-Unix OS (I think they use `\` instead of `/`). I'd recommend to use `os.path.join(…)` instead:

```python
urlretrieve(url,  os.path.join('data', 'average_real_estate_prices_by_%s.csv' % by))
```

cuducos:
No need to change this now (just keep in mind for next time), but we recommend starting the branch name with your initials (mine are `ec-some-feature`, `ec` stands for Eduardo Cuducos). Gonna pull the branch and start the code review now, get back to you soon!

cuducos:
Code wise it looks good, except the details about `os.path.join` I mentioned in a in-line comment.

Once you fetch the database it is important to translate the headers to English — this is still pending in your PR).

But my main question is about the granularity of the data. I'm not sure how far we can go with such “wide” price zones (by state). I think it would be more reliable to work with more granular average prices by street, postal code or at least neighborhood — but I'm not sure where to get this data though.

But I'm open for *pair programming* to explore some possibilities ; )

Lrcezimbra:
I think this PR can be closed without merge because #100 overwrites this, solving the issue of granularity.

PR #30, opened 2016-08-31
Add Telegram info on readme

brunopazzim:


cuducos:
I'd reinforce to the *hue hue* (aka Brazilian people) that the conversation is in English.

And reinforce to English speakers that *in spite of* being mostly Brazilians we're an international group and handle the conversation in English.

cuducos:
What do you think about 7c22454?

PR #35, opened 2016-09-02
Script to fetch receipt images

### LINKED ISSUE #33: Backup pictures from all receipts
It is vital for the project have a way of accessing all receipts, from any reimbursement since the first available and not depend from Chamber of Deputies.

Besides having proofs for legal reports, its useful for offline analyses. #32 is one I think about; doing OCR for generating new structured data is another.

Here's a function that, based on a record from quota datasets, returns the picture URL from the Chamber of Deputies' website:

```python
def document_url(record):
    return 'http://www.camara.gov.br/cota-parlamentar/documentos/publ/%s/%s/%s.pdf' % \
        (record['applicant_id'], record['year'], record['document_id'])
```

cuducos:
This is a program to download receipts to a local directory (fix #33).

```console
$ python src/fetch_receipts.py
    This script downloads the receipt images from the Lower House server.

    Be aware that downloading everything might use more than 1 TB of disk
    space.  Because of that you have to specify one `target` directory (where
    to save the files) and optionally you can specify with `--limit` the number
    of images to be downloaded.

    If the `target` directory exists and already has some saved receipts,
    these receipts will not be downloaded again (and they will not count when
    using `--limit` either).

    In other words, if you already have 42 receipts in your target folder,
    running the command with a limit of 8 will end up in a directory with 50
    files: the 42 you already had and 8 freshly downloaded ones.


positional arguments:
  target                Directory where images will be saved.

optional arguments:
  -h, --help            show this help message and exit
  -l LIMIT, --limit LIMIT
                        Limit the number of receipts to be saved
```

It will download the files to a different target than `data/` because it tends to use a huge amount of disk space (maybe more than 1TB).

That said  people might want to point it to a external volume (using `target` positional argument) and/or to limit the amount of receipts to be downloaded (using `--limit` optional argument).

It uses a new external library (added to `conda_requirements.txt`) called [humanize](http://github.com/jmoiron/humanize). You can manually install it with `pip install humanize` if you want.

Finally I recommend everyone that has a spare TB at home to save them until we can raise money to afford a virtual drive for the project with that amount of free space.

Using it to download everything to `/Volumes/Narnia/Serenata`:

```console
$ python src/fetch_receipts.py /Volumes/Narnia/Serenata
```

Using it to download 10,000 receipts to `/Volumes/Narnia/Serenata`:

```console
$ python src/fetch_receipts.py --limit 10000 /Volumes/Narnia/Serenata
```

### DIFF
--- a/src/fetch_receipts.py
+++ b/src/fetch_receipts.py
@@ -1,0 +1,177 @@
+import os
+import re
+import sys
+from argparse import ArgumentParser, RawTextHelpFormatter
+from urllib.error import HTTPError
+from urllib.request import urlretrieve
+
+from humanize import naturalsize
+import numpy as np
+import pandas as pd
+
+
+class Receipts:
+
+    BASE_DIR = os.path.dirname(os.path.dirname(os.path.abspath(__file__)))
+    DATA_DIR = os.path.join(BASE_DIR, 'data')
+    REGEX = r'^[\d-]{11}(current-year|last-year|previous-years).xz$'
+
+    def __call__(self):
+        return self.receipts
+
+    def datasets(self):
+        """List generator with the full path of each CSV/dataset."""
+        for file_name in os.listdir(self.DATA_DIR):
+            if re.compile(self.REGEX).match(file_name):
+                yield os.path.join(self.DATA_DIR, file_name)
+
+    def all(self):
+        """
+        List generator with Receipt named tuples containing the path of the
+        receipt image (to be used when saving it, for example) and the URL of
+        the receipt at the Lower House servers.
+        """
+        dtype = {
+            'document_id': np.str,
+            'congressperson_id': np.str,
+            'congressperson_document': np.str,
+            'term_id': np.str,
+            'cnpj_cpf': np.str,
+            'reimbursement_number': np.str
+        }
+        for dataset in self.datasets():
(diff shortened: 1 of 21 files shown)


src/fetch_receipts.py
```diff
@@ -0,0 +1,175 @@
...
+        # save receipts
+        self.limit = limit
+        self.target = target
+        self.progress = {
+            'count': 0,
+            'size': 0,
+            'errors': list(),
+            'skipped': list()
+        }
+        for receipt in self.receipts:
```
mtrovo: I think the operation of downloading the receipts doesn't fit on the constructor of the class, maybe it's a good idea to separate it into a `run` method which is going to fetch the receipts and also a `print_report` which will print the error messages if there's something.

cuducos: That sounds like a good idea, thanks!

Do you wanna step in and send a PR to this branch? These days my workload is focused on Issue #34. Otherwise I'l catch up here later.

cuducos:
Done, @mtrovo! Way better now. Thanks for the advice.

mtrovo:
Hey @cuducos I made this change on a local branch and added some improvements to download images in parallel, can you take a look at #50.

PR #47, opened 2016-09-09
Solve typos

patrickelectric:


Irio:
Thanks for you contribution, @patrickelectric. For the next time, `.html` and .py` files are autogenerated, so re-saving `.ipynb`will bring few typos back.

PR #53, opened 2016-09-11
Prevent unnecessary downloads/uploads of datasets and documentation

Irio:
Also adds the [datasets documentation](https://github.com/datasciencebr/serenata-de-amor/pull/25) as one of the datasets to be fetched by default.

### DIFF
--- a/develop/2016-08-08-im-translate-dataset.html
+++ b/develop/2016-08-08-im-translate-dataset.html
@@ -12044,7 +12044,7 @@ Name: 0, dtype: object</pre>
 </div>
 <div class="inner_cell">
 <div class="text_cell_render border-box-sizing rendered_html">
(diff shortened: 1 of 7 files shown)


cuducos:
Seems good to merge to me — I just haven't re-run all notebooks because it would take hours to recreate the CSV (some of the use the datasets in pt-br, not the ones from S3). But I don't think it's a problem.

cuducos:
Not necesarily linked to this PR, but as we are updating file names, what about fixing that?

```console
$ python src/translation_table.py
Traceback (most recent call last):
  File "src/translation_table.py", line 274, in <module>
    file_handler.write('\n'.join(markdown()))
  File "src/translation_table.py", line 269, in markdown
    for index, contents in enumerate(zip(get_portuguese(), EN)):
  File "src/translation_table.py", line 217, in get_portuguese
    with open('data/2016-08-08-datasets-format.html', 'rb') as file_handler:
FileNotFoundError: [Errno 2] No such file or directory: 'data/2016-08-08-datasets-format.html'
```

`src/translation_table.py` expects `datasets-format.html` to be prefixed with the date but it isn't. We can add the prefix or remove the date from this script.

Irio:
For now (422bf5d) I'm adding the date everywhere. This will break if trying to run with more recent files, but it's a problem yet to be present. Let's fix it when really becomes a problem.

Issue #29, opened 2016-08-31
Negative expenses on the dataset


andrepinho:
[@Irio 's descriptive analysis](https://github.com/datasciencebr/serenata-de-amor/blob/master/develop/2016-08-13-im-descriptive-analysis.ipynb) shows that we have documents with negative `net_value` in `2016-08-08-last-year.xz` dataset. All of them are flight tickets issued but not used by the congressperson.

My hypothesis is: for each of those documents there must be a corresponding document with a positive `net_value` where the flight ticket was first issued.

It can easily be proved since `document_number` should be the same for both documents.
One thing to keep in mind is that flight companies usually don't offer full reimbursement. That means the negative values will often be smaller than their positive counterpart (well, they're negative, but you get what I mean).

I think it's important to validate this because negative values without a corresponding positive one can hurt the accuracy of calculations regarding `net_value`.

c4rl0sm3nd3s:
Hi there,

I asked at Chamber of Deputies about what really happens to the "bilhete de compensação", because it is not clear that the negative value is really returned to the House or if it is somehow a open credit, so the Deputy could use later that amount for another reason (since it is already paid).

the protocol number is BAE9103619929 at http://www2.camara.leg.br/participe/fale-com-a-ouvidoria/fale_conosco_acompanhar_mensagem

I update here when I have some feedback from them.
thanks

andrepinho:
Hi @c4rl0sm3nd3s

Thanks for asking the Chamber and sharing with us. AFAIK, we really don't have an official statement regarding this issue. We've been just assuming that the congress person cancels the flight and then returns the money he received from the Flight Company.
Usually a reimbursement is less money than the purchase due to cancellation costs but I've seen it being higher. Maybe due to currency exchange rate? Not sure.
Have you seen my [PR](https://github.com/datasciencebr/serenata-de-amor/pull/66) about it? It's still a work in progress, but it show at least a little of how those flight reimbursements look like in the data.

c4rl0sm3nd3s:
I was doing some research and I found this 
![screen shot 2016-10-20 at 12 35 45 am](https://cloud.githubusercontent.com/assets/658836/19546374/4639bbe8-965d-11e6-80df-39e81d6bf309.png)

As you can see, all results has the same 'document_number' , in my sql db I use the same name as the xml, so column it is the 'txtNumero'. 

Maybe all the results should be used as one receipt, sum all and the result, if it is a positive value will be the 'net_value'.  

[...]

Issue #32, opened 2016-09-01
Detect multiple reimbursements using the same receipt


Irio:
For instance, we detected the same receipt being used in 2 distinct reimbursements:
http://www.camara.gov.br/cota-parlamentar/documentos/publ/2437/2015/5645173.pdf
http://www.camara.gov.br/cota-parlamentar/documentos/publ/2437/2015/5645177.pdf

cuducos:
If you remeber by heart (otherwise I look for it in the `.ipynb`): Do they have exactly the same `document_number` in the dataset? Or this number was mocked?

Just asking because if the real `document_number` differs (image vs dataset) we'll have to rely on OCR and stuff. If they are the same I think it's easier to spot.

Irio:
@cuducos In this example they have distinct `document_id`'s.

Before going to OCR, I'd try [SIFT](https://en.wikipedia.org/wiki/Scale-invariant_feature_transform), which I believe is much faster since does not depend on a vocabulary of words, just plain linear algebra.

cuducos:
Sound great. SIFT is new for me but looks like something very effective for this kind of stuff. Awesome!

paulo-raca:
I feel like SIFT is great for find similar stuff (e.g., receipts with the same layout), but is probably not going to be a good option to decide if 2 receipts are the same or not.

Irio:
Check the paper "Region Duplication Forgery Detection Technique Based on SURF and HAC" for references (https://sci-hub.cc/ is your friend). [Here](https://github.com/Irio/photoshopped-or-not/blob/master/develop/2016-06-16-im-sift-surf-region-duplication.ipynb)'s an example of Python code to run SIFT.

weslleymberg:
Came across 2 examples where 2 distinct reimbursements have the same `document_number`, but do not have the same receipt.

On the first one the value that is presented as the `document_number` is acctualy the congressperson's subscription number on the water company that issued the bills.

Here are the `document_id`s: [5886345](http://www.camara.gov.br/cota-parlamentar/documentos/publ/3074/2015/5886345.pdf) and [5886361](http://www.camara.gov.br/cota-parlamentar/documentos/publ/3074/2015/5886361.pdf).
And the `document_number` is 0010100910378000. You can see this is the same number that is in the field "Inscrição" on both documents.

[...]

PR #66, opened 2016-09-15
Analysing expenses with negative values

andrepinho:
I'm trying to figure out how these expenses with negative net_value impact on last year's data-set.
I tried to test the hypothesis on this issue #29 but it's not going to be as simple as I thought it would.

TL;DR
Looking only at last year's data-set:
- There's 17646 of those expenses.
- Looking for those 17646 `document_number` in positive expenses finds 13458 matches;
- When trying to isolate the negative values by removing the ones with positive matches I got 
  4458 expenses (only 4403 unique).

Then looking at previous year's data-set:
- 6655 matches for the 4458 (negative values without a positive counterpart) through `document_number`;
- From those, 17 have negative value;

So... it's a mess. I'll appreciate any insight that might explain those numbers as well as ideas on how to enhance this analysis in general.

### DIFF
--- a/develop/2016-08-31-ap-exploring-negative-net-values.html
+++ b/develop/2016-08-31-ap-exploring-negative-net-values.html
@@ -1,0 +1,12667 @@
+<!DOCTYPE html>
+<html>
+<head><meta charset="utf-8" />
+<title>2016-08-31-ap-exploring-negative-net-values</title>
+
+<script src="https://cdnjs.cloudflare.com/ajax/libs/require.js/2.1.10/require.min.js"></script>
+<script src="https://cdnjs.cloudflare.com/ajax/libs/jquery/2.0.3/jquery.min.js"></script>
+
+<style type="text/css">
+    /*!
+*
+* Twitter Bootstrap
+*
+*/
+/*!
+ * Bootstrap v3.3.6 (http://getbootstrap.com)
+ * Copyright 2011-2015 Twitter, Inc.
+ * Licensed under MIT (https://github.com/twbs/bootstrap/blob/master/LICENSE)
+ */
+/*! normalize.css v3.0.3 | MIT License | github.com/necolas/normalize.css */
+html {
+  font-family: sans-serif;
+  -ms-text-size-adjust: 100%;
+  -webkit-text-size-adjust: 100%;
+}
+body {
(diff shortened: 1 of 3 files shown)


pingfreud:
@andrepinho , any further advances?

This looks nasty.

andrepinho:
@pingfreud No real advances yet. I actually didn't look at the analysis since I opened the PR (shame on me). Been thinking about it though: one idea I have is to see if there's any `document_number` showing up in more than 2 expenses. That might explain some of these number while introducing the question: How can there be the same flight more than twice?

Will give it a try soon enough.

weslleymberg:
I think `document_number` alone is not very reliable because it is very prone to human error. I've talked something about it [here](https://github.com/datasciencebr/serenata-de-amor/issues/32#issuecomment-254592094) and [here](https://github.com/datasciencebr/serenata-de-amor/issues/32#issuecomment-254618059) on issue #32.

Maybe adding other columns such as `applicant_id` or `congressperson_id` (or any other column you find interesting) on the filter can help you find more reliable matches.

Another thing to try is to get a list with all unique tiquets, try and remove the prefix and search by the ticket's number alone with something like `df[df['document_number'].str.contains("<ticket number>")]` (not sure about the syntax)

I think this can help you with these questions:

> Those remaining 4.188 are negative expenses without a corresponding positive one?;

> The document numbers are messed up? Maybe something like: "Bilhete: 957-2117.270689" for the negative and "Vôo: 957-2117.270689" for the positive;

Hope this helps =)

Irio:
@andrepinho Before we can merge it, please finish your analysis with the following changes:

- [ ] Explain the reason you're doing the analysis in the notebook itself. The explanation may link to a GitHub issue, but the notebook should be self-contained.
- [ ] Explain the conclusions you had with the analysis. You may check [this example](https://github.com/datasciencebr/serenata-de-amor/blob/master/develop/2016-11-13-irio-doublechecking-net-values.ipynb) before writing your own.
- [ ] Make sure `.html` and `.py` files are up to date with your `.ipynb`.

andrepinho:
@Irio I'm closing this PR because it's out of date with the latest findings. The rework on net_values actually solves the issue that this analysis tries to shed light upon.

PR #50, opened 2016-09-10
Multiprocessing to the script that fetches receipt images

mtrovo:
Follow up to #35 adding support for download images in parallel and executing the download outside the constructor of the class FetchReceipts.

PR #86, opened 2016-10-18
Added docstrings to the new methods

lscasanova:
Added the sugestions made by @cuducoson on PR #50


cuducos:
Hi @lscasanova, I think it's better to send this PR to @mtrovo's branch, he merges it there and then from there we merge to `master` — is that ok?

lscasanova:
Ok!

Issue #37, opened 2016-09-03
Pack common tools in a specific module


cuducos:
While coding, planning to code or code reviewing I'm starting see a lot of repetition. My intention is to create a new module (`tools/` maybe) with tools to be used by anyone in other scripts (setting up scripts from `src/`, exploring and analyzing data in `develop/` etc.).

Anyone is free to join me in this, we can schedule and pair as this has no urgency at all.

And anyone can contribute listing functions that could be inside this module. I get started:
- list path to the CSV (`.xz`), i.e. the existing datasets
- load translations
- list all record (with or without filters) from the datasets
- …

Irio:
Been having this need for some time already, but just thought about an issue: Jupyter Notebooks are supposed to be totally reproducible; analyses we have, if ran again, should generate the exact same results. e.g. If I use a function `load_dataset('all')` in my analysis, its output might change when the function receives a patch.

Personally believe that having these functions will come for the better, even if bringing this issue with them. If we start noticing real problems, we could always use [`inspect.getsourcelines`](https://docs.python.org/3/library/inspect.html#inspect.getsourcelines) inside the notebook.

cuducos:
I'd add to the list:

*

cuducos:
Fixed by [serenata-toolbox](https://github.com/datasciencebr/serenata-toolbox) (even if we have to actually use it in this repo, as in #143)

Issue #15, opened 2016-08-21
Get names of immediate relatives of each deputy and senator


Irio:
Super useful for detecting different forms of nepotism. Better if we can get until third degree.

jvsl:
@Irio, Where Can I get those information?

cuducos:
Maybe Wikipedia, maybe Facebook… we have to be creative on this point…

Irio:
A good start is collecting who politicians declare on their own Facebook profiles.

Let's take Renato Molling, the RS federal deputy who spent more from the Quota for Exercising Parliamentary Activity last year. He has a personal Facebook profile and [lists two people as his family members](https://www.facebook.com/deputadorenatomolling/about?section=relationship&pnref=about), Larissa Molling and Vinicius Molling. In this specific case, listed relatives links to no Facebook profile directly, but searching for their names bring them as top results.

I believe [Facebook API](https://developers.facebook.com/) has ways of searching for people by their names and also returns the list of family members from the profile.

@jvsl

jvsl:
@Irio @cuducos , I was looking for a free service like this: https://www.myheritage.com.br/, but I didn't find. I think wikipedia and facebook  are a good start as cuducos said.

cuducos:
> P.S. Even if this method brings just a short list of politicians and relatives, we can give a high level of trust on this information, since it's auto declared.

cuducos:
Kudos for that @irio:

> P.S. Even if this method brings just a short list of politicians and relatives, we can give a high level of trust on this information, since it's auto declared.

jvsl:
I've been trying to use graph API to get family from  deputy and senator profiles, but I didn't have success. The API requires access token to get these information. The user token is obtained log in on facebook. Thus I'm just allowed to get my own information from facebook. So, I'll try to use wikipedia.

jvsl:
If someone wants to try using facebook, this link will help:
https://developers.facebook.com/tools/explorer/

cuducos:
> I've been trying to use graph API to get family from deputy and senator profiles, but I didn't have success. 

I'm sorry to hear that, @jvsl — we already knew that, and it was documented inthe [`README`](https://github.com/datasciencebr/serenata-de-amor#api). I don't want to sound like a dick, or even like mommy saying _I've told you_, but is there a way to make it clearer so people start off looking for alternatives?

[...]

PR #93, opened 2016-10-21
Create script to scrape the parents of congresspeople

braunmagrin:
The script will fetch the data from the congress website and then it
will create a file in data/congressperson_relatives.xz with 3 columns,
namely 'congressperson_id', 'relationship' and 'relative_name'. This
file will contain the names of the parent(s) of the congressperson.
The script will also create another file called
data/congressperson_relatives_raw.xz that will contain the original
string obtained from the site in case the script could not properly
split the father's and mother's name.

Maybe it'll be better to generate the files and then save them as you
did with the other files, because the script currently does more than
a thousand requests and it takes quite a long time.

### DIFF
--- a/src/get_family_names.py
+++ b/src/get_family_names.py
@@ -1,0 +1,152 @@
+# coding: utf-8
+
+import os
+import datetime
+import requests
+import numpy as np
+import pandas as pd
+from bs4 import BeautifulSoup
+
+PROCESSED_DATA_PATH = os.path.join('data', 'congressperson_relatives.xz')
+RAW_DATA_PATH = os.path.join('data', 'congressperson_relatives_raw.xz')
+DATE = datetime.date.today().strftime('%Y-%m-%d')
+
+write_csv_params = {
+    'compression': 'xz',
+    'encoding': 'utf-8',
+    'index': False}
+
+
+def format_string(s):
+    return s.strip().replace(':', '')
+
+
+def extract_contents_from_div(div):
+    return list(map(format_string,
+                    filter(lambda x: '\n' not in x,
+                           div[0].strings)))
+
+
+def convert_to_dict(contents):
+    return dict(zip(contents[1:-2:2],
+                    contents[2:-1:2]))
+
+
+def is_single_word(s):
+    return (' ' not in s)
+
+
+def fix_when_theres_a_single_surname_after_the_split(names):
+    pairs = [(i, name) for i, name
+             in enumerate(names) if is_single_word(name)]
+    while pairs:
+        i, name = pairs.pop(0)
+        if i == 0:
+            continue
+        names_to_join = names[i-1:i+1]
+        name = ' e '.join(names_to_join)
+        for n in names_to_join:
+            names.remove(n)
+        names.append(name)
+        pairs = [(i, name) for i, name
+                 in enumerate(names) if is_single_word(name)]
+    return names
+
+
+def split_names(s):
+    names = s.split(' e ')
+    names = fix_when_theres_a_single_surname_after_the_split(names)
+    return names
+
+
+def flatten_parents(df):
+    result = []
+    dict_df = df.to_dict(orient='records')[0]
+    for name in dict_df['parents_list']:
+        result.append({'id': dict_df['id'],
+                       'relative_name': name,
+                       'relationship': 'son_of'})
+    return pd.DataFrame(data=result)
+
+
+def get_all_congress_people_ids():
+    ids_series = [read_csv(name)['congressperson_id']
+                  for name in ['current-year', 'last-year', 'previous-years']]
+    return list(pd.concat(ids_series).unique())
+
+
+def read_csv(name):
+    return pd.read_csv('data/2016-08-08-%s.xz' % name,
+                       parse_dates=[16],
+                       dtype={'document_id': np.str,
+                              'congressperson_id': np.str,
+                              'congressperson_document': np.str,
+                              'term_id': np.str,
+                              'cnpj_cpf': np.str,
+                              'reimbursement_number': np.str})
+
+
+def write_formatted_data(df):
+    people_with_two_or_less_parents = (
+        df
+        [df.parents_list.apply(len) <= 2]
+        [['id', 'parents_list']])
+
+    final = (
+        people_with_two_or_less_parents
+        .groupby('id')
+        .apply(flatten_parents)
+        .reset_index(drop=True))
+
+    final.rename(inplace=True, columns={'id': 'congressperson_id'})
+    final.to_csv('-'.join(DATE, PROCESSED_DATA_PATH), **write_csv_params)
+
+
+def write_raw_data(df):
+    people_with_more_than_two_parents = (
+        df
+        [df.parents_list.apply(len) > 2]
+        [['id', 'Filiação']])
+
+    people_with_more_than_two_parents.rename(inplace=True, columns={
+        'id': 'congressperson_id',
+        'Filiação': 'parents'})
+
+    people_with_more_than_two_parents.to_csv('-'.join(DATE, RAW_DATA_PATH),
+                                             **write_csv_params)
+
+
+def main():
+    url = 'http://www2.camara.leg.br/deputados/pesquisa/layouts_deputados_' \
+          'biografia?pk={}'
+
+    ids = get_all_congress_people_ids()
+
+    global dicts
+    total = len(ids)
+    for i, id in enumerate(ids):
+        id = str(id).replace('\n', '').strip()
+        print('{} - {:.2f}%'.format(id, i / total * 100))
+        try:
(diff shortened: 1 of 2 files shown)


braunmagrin:
This is related to the issue #15 .

src/get_family_names.py
```diff
@@ -0,0 +1,152 @@
...
+    url = 'http://www2.camara.leg.br/deputados/pesquisa/layouts_deputados_' \
+          'biografia?pk={}'
+
+    ids = get_all_congress_people_ids()
+
+    global dicts
+    total = len(ids)
+    for i, id in enumerate(ids):
+        id = str(id).replace('\n', '').strip()
+        print('{} - {:.2f}%'.format(id, i / total * 100))
```
cuducos: You might use `end='\r'` as an argument of `print` to make each print replace the previous one in the Terminal (so instead of printing thousands of line, it's like an update of the script stdout in the line just above the terminal cursor where the command was entered). I use it in other scripts around here. I just don't use it when it's an error or warning.

Anyway… this is just an idea, not something that has to be done — it's totally up to you : )

src/get_family_names.py
```diff
@@ -0,0 +1,152 @@
...
+    df = pd.DataFrame(data=dicts)
+    df = df[df['Filiação'].notnull()]
+    df = df[~(df['Filiação'].str.contains('Escolaridade'))]
+    df['parents_list'] = df['Filiação'].apply(split_names)
+
+    write_formatted_data(df)
+    write_raw_data(df)
+
+
+dicts = []
```
braunmagrin: I've done it this way so I could access the intermediate data from the `main` function when running the script interactively (with `ipython -i src/get_family_names.py`). I can either make a comment in there to make this explicit or remove it and stop using `global`. What do you think?

src/get_family_names.py
braunmagrin: Great! I wasn't aware of this option. I'll definitely do it.

src/get_family_names.py
```diff
@@ -0,0 +1,152 @@
...
+        [['id', 'parents_list']])
+
+    final = (
+        people_with_two_or_less_parents
+        .groupby('id')
+        .apply(flatten_parents)
+        .reset_index(drop=True))
+
+    final.rename(inplace=True, columns={'id': 'congressperson_id'})
+    final.to_csv('-'.join(DATE, PROCESSED_DATA_PATH), **write_csv_params)
```
braunmagrin: Thanks for the detailed feedback and suggestions! It does make sense. That was a last minute change and I guess I didn't test it afterwards :(

src/get_family_names.py
```diff
@@ -0,0 +1,152 @@
...
+
+
+def get_all_congress_people_ids():
+    ids_series = [read_csv(name)['congressperson_id']
+                  for name in ['current-year', 'last-year', 'previous-years']]
+    return list(pd.concat(ids_series).unique())
+
+
+def read_csv(name):
+    return pd.read_csv('data/2016-08-08-%s.xz' % name,
```
braunmagrin: Well spotted. I guess that it would be easier to use if it would simply get the latest files. Agree?

braunmagrin:
I've fixed some issues and will work on the other ones later this week. I'll be glad if you could give me your opinion on the two remaining issues.

src/get_family_names.py
cuducos: **If** this a matter of facilitating debug, I'd remove line 150 and stop using `global`…

src/get_family_names.py
cuducos: Ooops… useless comment above, you've got it covered in 5e705f9. My bad.

src/get_family_names.py
cuducos: Totally ; )

cuducos:
Sure thing. I guess I just replied to both of them now.

Another thing: you can add a description and brief instructions of this script updating [the Scripts part of `CONTRIBUTING.md`](https://github.com/datasciencebr/serenata-de-amor/blob/master/CONTRIBUTING.md#suppliers-information-cnpj).

Hell yeah 🤘

src/get_family_names.py
```diff
@@ -1,15 +1,17 @@
 # coding: utf-8
```
cuducos: As the project is based on Python 3, UTF-8 is the default, no need for this `coding` here ; )

src/get_family_names.py
```diff
@@ -70,13 +72,23 @@ def flatten_parents(df):
...
     ids_series = [read_csv(name)['congressperson_id']
                   for name in ['current-year', 'last-year', 'previous-years']]
     return list(pd.concat(ids_series).unique())
 
 
+def find_latest_date():
+    date_regex = re.compile('\d{4}-\d{2}-\d{2}')
+
+    matches = (date_regex.findall(f) for f in os.listdir(DATA_DIR))
+    dates = (l[0] for l in matches if l)
```
cuducos: Not sure if this is the best approach because we could have other documents newer than the trio `current-year`/`last-year`/`previous-years` and that would fool us into getting the wrong date to that trio. For example, in a directory like that:

```console
2016-10-12-congressperson_relatives.xz
2016-08-08-current-year.xz
...
```

`find_latest_date()` would return `2016-10-12`, but we won't be able to find `2016-10-12-current-year.xz`.

In [another script](https://github.com/datasciencebr/serenata-de-amor/blob/master/src/fetch_receipts.py#L20) I used this regular expression if it helps: `r'^[\d-]{11}(current-year|last-year|previous-years).xz$'` (11 number or dash characters, then one of the file names we're looking for). Not the most precise regular expression ever, but does the job…

Maybe grouping the date and fetching only the required group will be even better. For example:

```python
regex = r'^(?P<date>\d{4}-\d{2}-\d{2})(-)(current-year|last-year|previous-years).xz$'`
results = re.match(regex, '2016-08-08-current-year.xz')
results.group('date')  # '2016-08-08'
```

Anyway, just brainstorming here… ; )

braunmagrin: Makes sense. I'll have a look at how the other scripts are doing it (I had only looked at the `fetch_cnpj_info`) and use your suggestions to come up with a good way to load files.

braunmagrin:
Maybe after this is merged, I could refactor these functions to load data into a separate module. I guess it would be easier for anyone else to write new scripts with that sort of helper functions.

cuducos:
> I could refactor these functions to load data into a separate module

There's already an issue for this ; ) #37 — feel free to jump in!

src/get_family_names.py
josecostamartins: During the grupy-rp sprint, I had this idea: we could fallback to older dates in case the file doesn't exists, so the method find_newest_file would get the dates from newest to oldest returning the newest file.

```python
def find_newest_file(name):
    """
    Assuming that the files will be in the form of :
    yyyy-mm-dd-type_of_file.xz we can try to find the newest file
    based on the date, but if the file doesn't exist fallback to another
    date until all dates are exhausted
    """
    date_regex = re.compile('\d{4}-\d{2}-\d{2}')

    matches = (date_regex.findall(f) for f in os.listdir(DATA_DIR))
    dates = sorted(set([l[0] for l in matches if l]), reverse=True)
    for file_date in dates:
        current_filename = 'data/{}-{}.xz'.format(date, name)
        if os.path.isfile(current_filename):
            return current_filename

    return None
```

Then the read_csv function would need to be modified for something like this:

```python
def read_csv(name):
    filename = find_newest_file(name)
    if (filename is None):
        raise TypeError('could not find the \
                specified filename: {}'.format(filename))

    return pd.read_csv('data/{}-{}.xz'.format(date, name),
                       parse_dates=[16],
                       dtype={'document_id': np.str,
                              'congressperson_id': np.str,
                              'congressperson_document': np.str,
                              'term_id': np.str,
                              'cnpj_cpf': np.str,
                              'reimbursement_number': np.str})
```

cuducos: I like both solutions — it's up to you, @braunmagrin  ; )

cuducos:
Thanks @josecostamartins for the fix!

@braunmagrin: @anaschwendler and I added some lines to fix the file loader and now it's merged 🎉 

Many thanks for the PR — and enjoy your #caipyra tshirt (thanks to @rougeth)

braunmagrin:
@cuducos I guess we did the same fixes on the loader :(

cuducos:
Sure thing. Sorry about that — my fault, I should've told you that I was working on t to fix ; )
Anyway, great contribution!

braunmagrin:
No problems! Thanks for the patience and all the feedback.
Looking forward to work with you again @cuducos and @anaschwendler ;)

anaschwendler:
Thank you so much @braunmagrin ! It was a great contribution!
It was awesome to work with you, if we plan another sprint that you can participate i am sure that we can work together again :)

:tada:

PR #94, opened 2016-10-23
Update CONTRIBUTING.md

josiasmainardi:
Just replaced Jarba's by Jarbas', as the original name already have 's' in the end.


cuducos:
Tks @josiasmainardi ; )

PR #96, opened 2016-10-27
Create script to distribute cases between investigators

Irio:
Requires 2 private files: cases currently under investigation, and
volunteer investigators.

Check
https://datasciencebr.com/o-que-atualmente-acontece-com-den%C3%BAncias-feitas-na-opera%C3%A7%C3%A3o-serenata-de-amor-49d75c01b509
for more information about the investigation process (in Portuguese).

---

I don't have much experience documenting Python scripts. @cuducos could use some of your guidance or at least a review here.

### DIFF
--- a/src/distribute_cases.py
+++ b/src/distribute_cases.py
@@ -1,11 +1,12 @@
 import configparser
 from df2gspread import gspread2df as g2d
 import numpy as np
+from os.path import expanduser, isfile
 import pandas as pd
 from random import sample
 
 """
-This script requires the existence of 2 files:
+This script requires the existence of 3 files:
     * `data/volunteers.csv`. Private, containing emails of investigators.
         e.g.
 
@@ -23,7 +24,11 @@ This script requires the existence of 2 files:
         1000,1,"john@example.com,michael@example.com,hannah@example.com","john@example.com"
         1001,1,"john@example.com,michael@example.com,hannah@example.com","john@example.com"
         1002,1,"john@example.com,michael@example.com,hannah@example.com","john@example.com"
-        1050,1,"barbara@example.com,susan@example.com",
+        1050,2,"barbara@example.com,susan@example.com",
+
+    * `~/.gdrive_private`. Needed to access your Google Drive documents.
+        Check the following link for further instructions.
+        https://github.com/maybelinot/df2gspread#access-credentials
 """
 
 INVESTIGATOR_EMAIL_COLUMN = 'Email (é por ele que iremos manter contato)'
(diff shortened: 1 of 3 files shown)


src/distribute_cases.py
```diff
@@ -0,0 +1,74 @@
...
+        available_investigators = investigators - \
+            already_in_this_case - \
+            set(batch.keys())
+        random_investigators = sample(available_investigators, n)
+        for email in random_investigators:
+            batch[email] = row['case']
+    return batch
+
+def read_cases_document():
+    file_path = expanduser('~') + '/.gdrive_private'
```
cuducos: Probably this is going to crash in Windows (as it uses `\` instead of `/`), but [`os.path.join`](https://docs.python.org/3/library/os.path.html#os.path.join) is always safe (as it uses [`os.sep`](https://docs.python.org/3/library/os.html#os.sep) behind the scenes):

```python
file_path = os.path.join(os.path.expanduser('~'), '.gdrive_private')
```

src/distribute_cases.py
```diff
@@ -0,0 +1,74 @@
+import configparser
+from df2gspread import gspread2df as g2d
+import numpy as np
+from os.path import expanduser, isfile
+import pandas as pd
+from random import sample
+
```
cuducos: Disclaimer: 💅

Python style guide propose [a convention to organize the `import` section](https://www.python.org/dev/peps/pep-0008/#imports). In a nutshell first we have native modules, then modules installed via `pip`, and then your application's modules. And on top of that, `import` comes before `from … import …`.

```python
import configparser
from os.path import expanduser, isfile
from random import sample

import numpy as np
import pandas as pd
from df2gspread import gspread2df as g2d
```

But… well… again: 💅

src/distribute_cases.py
```diff
@@ -0,0 +1,74 @@
+import configparser
+from df2gspread import gspread2df as g2d
```
cuducos: We need to add `gspread2df` to `conda_requirements.txt`, don't we?

src/distribute_cases.py
```diff
@@ -0,0 +1,74 @@
...
+        1050,2,"barbara@example.com,susan@example.com",
+
+    * `~/.gdrive_private`. Needed to access your Google Drive documents.
+        Check the following link for further instructions.
+        https://github.com/maybelinot/df2gspread#access-credentials
+"""
+
+INVESTIGATOR_EMAIL_COLUMN = 'Email (é por ele que iremos manter contato)'
+
+def distribute_cases(cases, investigators):
```
cuducos: As you've requested, sir 🍸

```python
def distribute_cases(cases, investigators):
    """
    Matches investigator email addresses and case ID numbers.
    :param cases: Google Spreadsheet downloaded by `gspread2df.download` method.
    :param investigators: (set) set of investigator email addresses (as strings)
    :return: (dict) dictionary containing investigator email addresses as keys
    and case IDs and values
    """ 
    ...
```

Dos this matches what you had in mind?

cuducos:
Another idea: use `argparse` to show the big description as the output for `--help`:

```console
$ python src/distribute_cases.py --help

This script requires the existence of 3 files:
    * `data/volunteers.csv`. Private, containing emails of investigators.
        e.g.
        "Email (é por ele que iremos manter contato)"
        john@example.com
        michael@example.com
        hannah@example.com
        barbara@example.com
        susan@example.com
    * Spreadsheet hosted at Google Drive. Private, containing cases.
        e.g.
        document_id,case,investigators,investigators_working
        1000,1,"john@example.com,michael@example.com,hannah@example.com","john@example.com"
        1001,1,"john@example.com,michael@example.com,hannah@example.com","john@example.com"
        1002,1,"john@example.com,michael@example.com,hannah@example.com","john@example.com"
        1050,2,"barbara@example.com,susan@example.com",
    * `~/.gdrive_private`. Needed to access your Google Drive documents.
        Check the following link for further instructions.
        https://github.com/maybelinot/df2gspread#access-credentials
```

cuducos:
Closing this issue as we tend to not use this strategy anymore.

PR #100, opened 2016-11-02
Collect real estates rental prices

Lrcezimbra:
Collect real estates rental prices from [ZAP Imóveis](http://www.zapimoveis.com.br/busca-mapa). It solves the problem of granularity cited on #27 and #20. 

Now it's collecting data from Distrito Federal, but can easily get from other locations. 
The collected data is: 
`ID, coordinates, rent_rate, sale_rate, rental_price, sale_price, total_area, useful_area, neighborhood, dorms, subtype, suite, vacancies`

I removed some columns that I think it's not necessary for the project.

### DIFF
--- a/src/fetch_real_estate_prices.py
+++ b/src/fetch_real_estate_prices.py
@@ -22,7 +22,7 @@ def slice_quadrants(northeast, southwest, size=0.0625):
 def get_real_estates_from_quadrants(quadrants):
     estates = []
     url = 'http://www.zapimoveis.com.br/BuscaMapa/ObterOfertasBuscaMapa/'
-    for quadrant in quadrants:
+    for i, quadrant in enumerate(quadrants):
         search_params = {
             "CoordenadasAtuais": {
                 "Latitude": -15.7217174,
@@ -40,15 +40,41 @@ def get_real_estates_from_quadrants(quadrants):
             "TipoOferta":"Imovel"
         }
 
-        data = {'parametrosBusca': str(search_params)}
-        req = requests.post(url, headers=HEADERS, data=data)
-        estates_data = json.loads(req.text)
-        results = estates_data.get('Resultado')
+        results = get_results(url, {'parametrosBusca': str(search_params)})
+
         if results:
             results_not_fetched_amount = results.get('QuantidadeResultados') - 1000
             if results_not_fetched_amount > 0:
                 print('{} results aren\'t being fetched, decrease the quadrant size to get more results'.format(results_not_fetched_amount))
-            estates.extend(tuple(results.get('Imoveis')))
+            estates.extend(results.get('Imoveis'))
+        print('Fetched {} out of {} ({:.2f}%)'.format(i+1, len(quadrants), (i+1) / len(quadrants) * 100), end='\r')
+    print('\nFetched {} real estates.'.format(len(estates)))
+
+    return estates
+
+def get_real_estates_details(real_estates):
+    details = []
+    all_ids = [e['ID'] for e in real_estates]
+    url = 'http://www.zapimoveis.com.br/BuscaMapa/ObterDetalheImoveisMapa/'
+    ids_per_request = 10
+    total = len(all_ids)
+
+    splitted_ids = [all_ids[x:x+ids_per_request] for x in range(0, total, ids_per_request)]
+
+    for i, ids in enumerate(splitted_ids):
+        results = get_results(url, {'listIdImovel': str(ids)})
+        details.extend(results)
+
+        ids_fetched = i*len(results)+1
+        print('Fetched {} out of {} ({:.2f}%)'.format(ids_fetched, total, ids_fetched/total*100), end='\r')
+    print('\nFetched {} real estates details.'.format(len(details)))
+
+    return details
+
+def get_results(url, data, headers=HEADERS):
+    request = requests.post(url, headers=HEADERS, data=data)
+    estates_data = json.loads(request.text)
+    return estates_data.get('Resultado')
 
 if __name__ == '__main__':
     from unittest import TestCase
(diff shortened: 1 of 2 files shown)


src/fetch_real_estate_prices.py
```diff
@@ -0,0 +1,126 @@
+import json
+import numpy as np
+import requests
+import os
+import pandas as pd
+from geopy.geocoders import GoogleV3
```
cuducos: I can't find this module `geopy`. Should we add it to `conda_requirements.txt`?

src/fetch_real_estate_prices.py
```diff
@@ -0,0 +1,126 @@
+import json
```
cuducos: Disclaimer: 💅

Python style guide propose [a convention to organize the `import` section](https://www.python.org/dev/peps/pep-0008/#imports). In a nutshell first we have native modules, then modules installed via `pip`, and then your application's modules. And on top of that, `import` comes before `from … import …`.

```python
import json
import os

import numpy as np
import pandas as pd
import requests
from geopy.geocoders import GoogleV3
```

But… well… again: 💅

src/fetch_real_estate_prices.py
```diff
@@ -0,0 +1,126 @@
...
+import numpy as np
+import requests
+import os
+import pandas as pd
+from geopy.geocoders import GoogleV3
+
+HTTP_HEADERS = {"User-Agent":"Mozilla/5.0 (Macintosh; Intel Mac OS X 10_8_5) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/32.0.1700.102 Safari/537.36"}
+OUTPUT_FILEPATH = os.path.join('data', 'real_estates_prices.csv')
```
cuducos: Your PR's made me realize we have been using a kind of standard for the `/data` that was undocumented. We are prefixing all files with date (`YYYY-MM-DD-my-dataset.xz`) and we are using `.xz` ([lzma](https://docs.python.org/3/library/lzma.html)) compression.

Would you mind updating the naming and format of the output file?

src/fetch_real_estate_prices.py
```diff
@@ -0,0 +1,126 @@
...
+LOCATION = 'Distrito Federal, Brasil'
```
cuducos: I'm wondering: how would this work for fetching data from all states?

src/fetch_real_estate_prices.py
```diff
@@ -0,0 +1,126 @@
...
+
+def slice_quadrants(northeast, southwest, size=0.0625):
```
cuducos: If we're dealing with tons of data, using generators might be a wise choice:

```python
def slice_quadrants(northeast, southwest, size=0.0625):
    for lat in np.arange(southwest['lat'], northeast['lat'], size):
        for long in np.arange(southwest['lng'], northeast['lng'], size):
            yield ((str(lat), str(long)), (str(lat + size), str(long + size)))
```

cuducos:
One more thing: you could document the purpose and usage of the script in `CONTRIBUTING.md` ; )

Lrcezimbra:
@cuducos thanks for the code review and tips. I'll work on this improvements.

src/fetch_real_estate_prices.py
Lrcezimbra: Already is in `conda_requirements.txt` line 2 `geopy==1.11.0`

src/fetch_real_estate_prices.py
Lrcezimbra: Fixed on 4e085d3

src/fetch_real_estate_prices.py
Lrcezimbra: Change to a list in b0bcc02 to support multiple locations.

src/fetch_real_estate_prices.py
Lrcezimbra: Good idea, implemented in b501c93.

src/fetch_real_estate_prices.py
Lrcezimbra: I was in doubt about how to organize the imports, but didn't search about. Thanks for the tip.

Fixed in 0b48e05. But noting now, requests is a native module, correct? I didn't pay attention to this.

Lrcezimbra:
I pushed some commits. Still missing:

- [ ] docstrings
- [ ] CONTRIBUTING.md
- [ ] lines too long

And I'm adding threads to run the script faster.
- [ ] Threads

I'll work on this asap.

src/fetch_real_estate_prices.py
cuducos: [`resquests`](http://docs.python-requests.org/en/master/) isn't native but I guess it's installed by default in Anaconda Python. The native module is [`urllib.request`](https://docs.python.org/3.5/library/urllib.request.html#module-urllib.request) — easy to mess up, isn't it? ; )

cuducos:
That TODO list, and the improvements, are awesome! Congrats and thank you!

Let me know if you need some help and I'd be happy to pair ; )

Lrcezimbra:
I've been kind of busy in the lasts weeks. Next week I'll finish the pull request.

Lrcezimbra:
Add docstring in b92609f. My English isn't very good, so if my docstrings aren't good, please let me know that I try to improve.

I think this latest commits solves all the fixes and improvements cited by @cuducos

cuducos:
For some reason I completely forgot to get back to this PR (probably because real state price was out of our first hypothesis)!

Sorry @Lrcezimbra!

Before merging I'd like to see two minor changes:

* Can you update `CONTRIBUTING.md` too avoid conflicts? If you want, I already [fixed conflicts and typos/wording](https://gist.github.com/a033205f36114b550591b61670ac076f)
* Can you extract `location` with `sys.argv` or `argparse`? So instead of editing the `.py` file, one just add the location in the terminal (eg `$ python src/fetch_real_estate_prices.py "Araguari, Minas Gerais"`)

Lrcezimbra:
I'll work on it.

cuducos:
Closed by inactivity – surely anyone is free to jump in and take this great contribution further ; )

PR #106, opened 2016-11-05
Added utils.py

leportella:
Added function to find the sum of value considering unique values of an aggregator


cuducos:
Related to Issue #37

Issue #98, opened 2016-10-31
Develop an algorithm to give the probability of a certain partner be a relative of the congressperson 


cuducos:
We have all the names of the congresspeople and [the name of their parents](https://github.com/datasciencebr/serenata-de-amor/pull/93). 

That said @anaschwendler and I were discussing today the possibility of having an algorithm that receives as an input:

* the name of a congressperson
* the name of the relatives
* the name of a partner of a company in which the congressperson has spent some public money

The algorithms would give us the probability of the following hypothesis: the partner and the congressperson are relatives.

We could balance more popular (e.g. Silva) and less popular family names (e.g. Sarney) with internal sources (we have thousands of full names in out dataset, including congresspeople and their parents) or with an external database (no ideia, but that should not be a big challenge).

PS: formally we don't have company partners in our dataset, but it's on our roadmap (and maybe the _development_ of this algorithm doesn't depend on that).

(@g4brielvs feel free to jump in!)

gabriel-almeida:
I was looking for some way to contribute with this project and this issue seems interesting, since I already solved a similar problem. Here are some ideas:

* This can be approached with the logistic regression learning algorithm, since it has a probabilistic interpretation. Also, scikit-learn has it and it is not overly complex to understand the final result.
* The features for the learning could be: the names in common, some measures of string similarity or a cleaver combination of both - need to think more about it. Related package: https://rawgit.com/ztane/python-Levenshtein/master/docs/Levenshtein.html
* The "negative part" of the dataset used to learn could be made with some random combinations of the available people (be aware that use all combinations may be infeasible)
* Some configurations of the learning algorithm depends on the use case: 
  * If it is a kind of triage, just to filter very bad cases (many false positives but high recall);
  * If it needs to make sure about the answer, even if it allows a few cases to pass (many false negative but high precision);

I might do this in the future, but feel free to steal those ideas :)

cuducos:
This is awesome! Many thanks, @gabriel-almeida!

I'm not sure I have the right skills to code that this quickly by myself, but surely this leaves this Issue way easier. Whoever wants to jump in, make yourself at home ; )

eldersantos:
Wikipedia provides a lot relationship between the politics, maybe it is a good source to scrape data

cuducos:
[...]

PR #107, opened 2016-11-05
Family name classifier

gabriel-almeida:
I managed to test the ideas that I suggested in #98 in these last days, so I'm pulling a baseline implementation (and getting frustrated with git's complexity). I could use the script from #93 to get the "congressperson_relatives" dataset and test the final learning algorithm. 

Some considerations:
* I don't know how to use pandas properly, so a made a really dirty work around just to get the data in a format that I could work with. The function "_get_relatives()" does this and it should be removed as soon as the dataset is ready.
* The algorithm was tested with "truncated" names, provided by the default "current year" dataset. The results should be more useful and realistic if the full name is used to train and predict, but it seems that they are not collected anywhere.
* The algorithm also needs way more data to get better. Any data with persons names and relatives names should be good, since it uses the frequency of surnames and first names to make some decisions (like remove composite names like "Pedro Paulo").
* I got a precision (not accuracy) score of about 80% with this final form of the code, but don't trust this result, a more careful analysis should be made. (https://en.wikipedia.org/wiki/Precision_and_recall)
* I tested the whole thing trying to find families in the provided data and this can be find in the "_demo()" function. There are some handpicked high probability (>80% certain) results below: 

```
Same Family: WALDIR MARANHÃO + BENJAMIN MARANHÃO
Same Family: DANILO CABRAL + JOSÉ CARLOS ALELUIA
	Reason: Maria do Amparo de Barros Cabral - Maria do Amparo Aleluia Costa
Same Family: IRAJÁ ABREU + JERÔNIMO GOERGEN
	Reason: Kátia Regina de Abreu - Clecia Regina Pizzolotto Goergen
Same Family: RICARDO BARROS + MARCUS PESTANA
	Reason: Barbara Cecily Netto Barros - Agostinho Pestana da Silva Netto
Same Family: AFONSO MOTTA + ÁTILA A. NUNES
	Reason: Livia Antunes da Motta - Tania Maria Antunes Nunes Pereira
Same Family: WASHINGTON REIS + FABIO REIS
	Reason: Sildes Reis de Oliveira - Jerônimo de Oliveira Reis

```
They aren't from the same family but it is a beginning! This should emphasis my point that it needs more data to evaluate common surnames and first names, and, of course, to not trust blindly in this kind of algorithm, especially if you have a database with the precise data that you need :)

### DIFF
--- a/src/family_name_classifier.py
+++ b/src/family_name_classifier.py
@@ -1,16 +1,17 @@
-import builtins
-
-import pandas as pd
-import numpy as np
 import itertools
 import unicodedata
 import re
 import collections
 import random
+
+import pandas as pd
+import numpy as np
 import sklearn.feature_extraction
 import sklearn.linear_model
 import sklearn.cross_validation
 import sklearn.externals.joblib
+
+
 random.seed(0)
 
 
@@ -69,8 +70,8 @@ class FamilyNameClassifier:
         :param name: string
         :return: is_first_name: boolean
         """
-        prob_firstname = 1.0 * self.first_name_counter[name] / self.total_first_names
-        prob_surname = 1.0 * self.surname_counter[name] / self.total_surnames
+        prob_firstname = self.first_name_counter[name] / self.total_first_names
+        prob_surname = self.surname_counter[name] / self.total_surnames
         return prob_firstname >= prob_surname
 
     def tokenize_surnames(self, name):
@@ -114,7 +115,7 @@ class FamilyNameClassifier:
 
         size_longest_name = max(len(tokens1), len(tokens2))
         if size_longest_name > 0:
-            features['match_proportion'] = 1.0 * len(common_names) / size_longest_name
+            features['match_proportion'] = len(common_names) / size_longest_name
 
         return features
 
@@ -208,16 +209,16 @@ class FamilyNameClassifier:
         """
         sklearn.externals.joblib.dump(self, file_name, compress=9)
 
-    @builtins.staticmethod
-    def load(file_name):
-        """
-        Loads a saved classifier.
 
-        :param file_name: string
-        :return: classifier
-        """
-        self = sklearn.externals.joblib.load(file_name)
-        return self
+def load_classifier(file_name):
+    """
+    Loads a saved classifier.
+
+    :param file_name: string
+    :return: classifier
+    """
+    self = sklearn.externals.joblib.load(file_name)
+    return self
 
 
 def _positive_data_iterator(relatives_dict):
--- a/src/family_name_classifier.py
+++ b/src/family_name_classifier.py
@@ -1,0 +1,315 @@
+import builtins
+
+import pandas as pd
+import numpy as np
+import itertools
+import unicodedata
+import re
+import collections
+import random
+import sklearn.feature_extraction
+import sklearn.linear_model
+import sklearn.cross_validation
+import sklearn.externals.joblib
+random.seed(0)
+
+
+class FamilyNameClassifier:
+    IGNORE_LIST = {'de', 'da', 'do', 'des', 'das', 'dos', 'e', 'dr', 'sobrinho', 'junior', 'neto', 'jr', 'filho'}
+    NOT_ACCEPTABLE_CHARACTERS = re.compile(r'[^a-z ]')
+
+    def __init__(self):
+        self.vectorizer = None
+        self.surname_counter = None
+        self.total_surnames = None
+        self.total_first_names = None
+        self.model = None
+
+    def tokenize(self, name):
+        """"
+        Simple tokenizer that normalizes and split a name by spaces.
+
+        :param name: string
+        :return: tokens: list
+        """
+
+        name = name.strip().lower()
+        name = unicodedata.normalize('NFD', name)
+        name = self.NOT_ACCEPTABLE_CHARACTERS.sub("", name)
+        return [n for n in name.split(' ') if n not in self.IGNORE_LIST and len(n) > 1]
+
+    def __count_names(self, relatives_dict):
+        """
+        Calculate some names statistics used internally.
+
+        :param relatives_dict: dict
+        :return:
+        """
+        self.surname_counter = collections.Counter()
+        self.first_name_counter = collections.Counter()
+
+        for son, parents in relatives_dict.items():
+            tokens = self.tokenize(son)
+            self.first_name_counter.update([tokens[0]])
+            self.surname_counter.update(tokens[1:])
+
+            for parent in parents:
+                tokens = self.tokenize(parent)
+                self.first_name_counter.update([tokens[0]])
+                self.surname_counter.update(tokens[1:])
+
+        self.total_surnames = sum([v for k, v in self.surname_counter.items()])
+        self.total_first_names = sum([v for k, v in self.first_name_counter.items()])
+
+    def is_first_name(self, name):
+        """
+        Says if a name (preprocessed by tokenize() function) is not a surname, based on the collected statistics.
+        Used internally to remove composite names.
+
+        :param name: string
+        :return: is_first_name: boolean
+        """
+        prob_firstname = 1.0 * self.first_name_counter[name] / self.total_first_names
+        prob_surname = 1.0 * self.surname_counter[name] / self.total_surnames
+        return prob_firstname >= prob_surname
+
+    def tokenize_surnames(self, name):
+        """
+        Returns a list of family names from a complete name string.
+
+        :param name: string
+        :return: surname_tokens: list
+        """
+        name_tokens = self.tokenize(name)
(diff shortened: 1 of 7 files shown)


src/family_name_classifier.py
```diff
@@ -0,0 +1,315 @@
+import builtins
```
cuducos: Disclaimer: 💅

Python style guide propose [a convention to organize the `import` section](https://www.python.org/dev/peps/pep-0008/#imports). In a nutshell first we have native modules, then modules installed via `pip`, and then your application's modules.

```python
import builtins
import collections
import itertools
import random
import re
import unicodedata

import numpy as np
import pandas as pd
import sklearn.cross_validation
import sklearn.externals.joblib
import sklearn.feature_extraction
import sklearn.linear_model
```

But… well… again: 💅

src/family_name_classifier.py
```diff
@@ -0,0 +1,315 @@
...
+import sklearn.feature_extraction
+import sklearn.linear_model
+import sklearn.cross_validation
+import sklearn.externals.joblib
+random.seed(0)
+
+
+class FamilyNameClassifier:
+    IGNORE_LIST = {'de', 'da', 'do', 'des', 'das', 'dos', 'e', 'dr', 'sobrinho', 'junior', 'neto', 'jr', 'filho'}
+    NOT_ACCEPTABLE_CHARACTERS = re.compile(r'[^a-z ]')
```
cuducos: Is there a specific reason to ignore special letters such as `ç`, `é`, and `ã`?

src/family_name_classifier.py
```diff
@@ -0,0 +1,315 @@
...
+        """"
+        Simple tokenizer that normalizes and split a name by spaces.
+
+        :param name: string
+        :return: tokens: list
+        """
+
+        name = name.strip().lower()
+        name = unicodedata.normalize('NFD', name)
+        name = self.NOT_ACCEPTABLE_CHARACTERS.sub("", name)
```
cuducos: Maybe if the script don't work with special chars we could some of them in ASCII characters:

```python
import unicodedata

value = 'fóãçêé'

for c in unicodedata.normalize('NFD', value):
    if unicodedata.category(c) != 'Mn':
        print(c)
```

Outputs:

```console
f
o
a
c
e
e
```

src/family_name_classifier.py
```diff
@@ -0,0 +1,315 @@
...
+        return [n for n in name.split(' ') if n not in self.IGNORE_LIST and len(n) > 1]
+
+    def __count_names(self, relatives_dict):
+        """
+        Calculate some names statistics used internally.
+
+        :param relatives_dict: dict
```
cuducos: What's the _shape_ of this dict? `{'congressperson name': ["congressperson's mon", "congressperson's dad"]}`?

src/family_name_classifier.py
```diff
@@ -0,0 +1,315 @@
...
+
+            for parent in parents:
+                tokens = self.tokenize(parent)
+                self.first_name_counter.update([tokens[0]])
+                self.surname_counter.update(tokens[1:])
+
+        self.total_surnames = sum([v for k, v in self.surname_counter.items()])
+        self.total_first_names = sum([v for k, v in self.first_name_counter.items()])
+
+    def is_first_name(self, name):
```
cuducos: Great idea! 👍

src/family_name_classifier.py
```diff
@@ -0,0 +1,315 @@
...
+        """
+        Says if a name (preprocessed by tokenize() function) is not a surname, based on the collected statistics.
+        Used internally to remove composite names.
+
+        :param name: string
+        :return: is_first_name: boolean
+        """
+        prob_firstname = 1.0 * self.first_name_counter[name] / self.total_first_names
```
cuducos: Do we need this `1.0 * `? The `/` operator always return a `float` in Python 3:

```python
>>> 1/3
0.3333333333333333
>>> 4/2
2.0
```

src/family_name_classifier.py
```diff
@@ -0,0 +1,315 @@
...
+                                                    names_iterator))
+        n_positives = dataset.shape[0] - n_negative_samples
+        target = [0] * n_negative_samples + [1] * n_positives
+
+        print("Training")
+        scoring = 'precision'
+        self.model = sklearn.linear_model.LogisticRegressionCV(n_jobs=-1, scoring=scoring)
+        self.model.fit(dataset, target)
+
+        # Debug part:
```
cuducos: ♥️

src/family_name_classifier.py
```diff
@@ -0,0 +1,315 @@
...
+    def save(self, file_name):
+        """
+        Saves the current (trained) object with scikit-learn Joblib method.
+
+        :param file_name: string
+        :return:
+        """
+        sklearn.externals.joblib.dump(self, file_name, compress=9)
+
+    @builtins.staticmethod
```
cuducos: Do we need `@builtins.staticmethod`? I think `@staticmethod` does the job… (if that's the case we can remove `import builtins` from the top of the file.

src/family_name_classifier.py
gabriel-almeida: Sorry, it is just force of habit, I always get a bug by an integer division.

src/family_name_classifier.py
gabriel-almeida: Could be an external function too: it is really just to simplify the load of a saved classifier.

src/family_name_classifier.py
gabriel-almeida: I tried to do this code in PEP 8, but seems like PyCharm forgot to warn me about this  (or I ignored it).

josecostamartins:
I did code review and ran the code in jupyter-notebook, everything works as Gabriel said, including the results.
As a baseline, implementation this is a good first-step.

Just a few considerations:
To run Gabriel's code I had to use #93 So, so there is a dependence.
Alson, I'm no expert, but considering the variables we have available (congressperson_id, relationship, father name, mother name), seems to me logistic regression appears not to be the best tool for this algorithm since the quantity of last names and their combinations might render it with a relatively low precision (not accuracy), maybe a heuristic with word_distance (NLTK implements this, but can be done with python directly) applied on last_names can handle interesting results.

gabriel-almeida:
I think it was me that messed up with the git, sorry about that, I'm not used to contribute with open source code.

I would like to explain my thoughts developing this, for documentation sake, that I did not explain in the pull request:

* I emphasis "precision" because the accuracy (overall hit) can be as high as you want, just increasing the negative part of the dataset: if the algorithm says that everyone is not from the same family it would be correct about 99% of the time. Analyze precision is more suitable in this case: if it has a precision of 85%, for example, then when it says that two persons are from the same family, then I have a confidence of 85% that it is true (maybe "chance" is a strong word here), but I am not caring about the cases that I let it go.

* I basically tried to left most of my decisions to the data, since the use of machine learning in the final decision and logistic regression was my algorithm of choice because it is just like Perceptron and Linear SVM: in the end, they are just a weighted sum of the features values plus an independent number (but logistic regression, in special, has an probabilistic interpretation). Of course, anything more complicated can be used, but the simplicity of those guys make it very easy to understand their final choice and they can perform as good as any other algorithm when the choice of the features are well made (especially when you have little data).

* Some combinations of features were tried until I could get a set of meaningful weights in the logistic regression, this code is just the end result. I tried to use string similarity (Python-Levenstein module) on the surnames but it appears to have confused the algorithm, maybe because this thing is more useful when you need just to account for typos: "Silva" and "Silvana" are very similar names but had very different meanings. Those kind of things can be used but you also will need to account for the lost of the frequency of a mistyped surname (a very important feature), how typos will be introduced in the training data and a possible increase in processing time.

* Any large set of Brazilian full names, even without parents information, could be used to improve the names statistics, since they are so important overall. (By the way, don't mind the logarithm black magic being used with the frequency of a surname, it is a common data mining trick when you have very few guys appearing in the bulk of cases, like the Silva surname).

Sorry for this thesis, but I hope that this clarifies my choices here. I might correct the commit problems, organize things a little more and maybe find more names to use.

src/family_name_classifier.py
cuducos: No worries at all… as I said, just 💅 hahaha…

src/family_name_classifier.py
cuducos: Do we need to remove `ç` and `é` and so on, or just punctuation, numbers? I suggested below we could convert `ç` to `c`, `é` to `e` and so on — but not sure if that means a better classifier… what do you think about it?

src/family_name_classifier.py
cuducos: Great. I'd add that _shape_ and your description to the __docstring__ to make it easier for readers and contributors ; )

gabriel-almeida:
I think I messed up with git trying to correct the brach. I might need some instructions or open another pull request... Sorry about that.

PR #109, opened 2016-11-06
Adding new functions to utils.py

juhhcarmona:


cuducos:
This new methods are good. I think we can better organize the data output. Let's pair on that to sort it out.

Issue #39, opened 2016-09-06
Find expenses in brothels


Irio:
Brothels will usually print receipts using companies with names and main activities more "morally acceptable", e.g. a restaurant. If we can gather a list of CNPJ's from the major brothels in Brazilian capitals, this could generate a very interesting report and news coverage.

Original idea by Pedro Vilanova.

cuducos:
I think we might gather great data with something as simples as a Google Form and a social media campaign (cc @tatianasb)

camilagss:
For Rio de Janeiro's data you can export [Bitchmaps' KML](https://www.google.com/maps/d/u/0/viewer?mid=1966bBklf94Uj5glEZGRigecFa60&hl=pt_BR).

ruivelocci:
Great data @camilagss. I think if we compare the time of spend with your presence on SESSION/CPI with that we can find "Strange Things".

rafaelalmeida:
What data does the receipts offer? @camilagss 's KML is great, but do most receipts have addresses, so we can geocode and match? How can I find some sample receipts?

Irio:
@rafaelalmeida Run `python src/fetch_datasets.py` and you'll get [a dataset](https://github.com/datasciencebr/serenata-de-amor/blob/c5439d947812ec40948eb445f7badf77dc6ad0e5/src/fetch_datasets.py#L31) with latitude/longitude of all companies in the receipts will be downloaded. It's `data/2016-09-03-companies.xz`.

https://github.com/datasciencebr/serenata-de-amor/pull/19

iGallina:
Politicians usually use the home delivery service and pay in cash for this kind of service.
Or get someone else to pay for them, which is more common.
We should focus on what is left out of the auditing or focus on what is being wrongly declared?

guilhermeprokisch:
Another source could be the
http://www.wikisexguide.com/

For example:
http://www.wikisexguide.com/wiki/S%C3%A3o_Paulo

guilhermeprokisch:
If I have time I will think how to create a webscraper for this website, they have the address of many places, we can check possibles "fake companies names", looking the address.

guilhermeprokisch:
Hey Guys. I think I can do a CSV with the http://www.wikisexguide.com/ containing the Name, Category, Localization, and Phone.  I will try to open a request Pull in some days.

guilhermeprokisch:
Hi Guys. 

[...]

PR #113, opened 2016-11-07
A search script to find suspicious places around a geolocalization

guilhermeprokisch:
Following the problem proposed in the #39 . In the lasts days, I was thinking how to solve this. I start to trying to create some web crawler for specialized sites of prostitutes, motels, sex shop and etc.. But a problem arise in this approach. Is hard to find specialized websites that contain huge information for all cities in Brazil. But doing this I noticed that almost all this places are in Google Maps with more soft descriptions. Like : "Night Clubs" or "Club of Adult Entertainment". I collected some of this keyword empirically and created a script to search on the Google Places API for places that  match a list of keywords given a geolocalization and return the closest place suspect.

### DIFF
--- a/src/suspect_place_around_search.py
+++ b/src/suspect_place_around_search.py
@@ -38,6 +38,8 @@ def search_suspect_place(lat, lng):
                         "Sex Shop",
                         "Massagem",
                         "Swinger clubs"
+                        "Adult Entertainment Store"
+                        "Modeling Agency"
                         "Sex Club",
                         "Gay Sauna",
                         "Strip Club"]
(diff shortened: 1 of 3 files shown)


src/suspect_place_around_search.py
```diff
@@ -0,0 +1,128 @@
...
+                - distance : The distance of suspect place and the given lat, lng
+                - address : The address  of suspect place
+                - phone : The phone of suspect place
+                - id : The Google Place ID of suspect place
+                - keyword : The Keyword that generate the place
+                            in Google Place Search
+
+    '''
+
+    GOOGLE_PLACES_API_KEY = 'SOME_GOOGLE_PLACE_API_KEY'
```
cuducos: Would you mind loading that from `config.ini` as in [`src/geocode_addresses.py`](https://github.com/datasciencebr/serenata-de-amor/blob/master/src/geocode_addresses.py#L18)?

Issue #21, opened 2016-08-23
Get social contract from Registro Livre


Lrcezimbra:
There is a project [Registro Livre](http://registrolivre.inf.br/) which is [open source](https://github.com/ThoughtWorksInc/registrolivre) that the idea is to provide the social contracts of companies that are available at the _Junta Comercial_. We could use these data to cross business partners or other informations.

Irio:
Hi @Lrcezimbra, thanks for your contribution. This is my first contact with the given project, and I can just see [10 companies listed in their site](http://registrolivre.inf.br/#/empresas). Do you know if they have more than that? If they just have this amount and don't plan to get information about more companies in the soon future, I find hard to understand how to connect both projects.

Lrcezimbra:
I think they intend to add more companies, but I'm not sure how the project is flowing. I had the idea to use this project because the Receita Federal and ReceitaWS I think don't have the data from the partners. So that would be an option if more companies are added.

pedrro:
Hi @Lrcezimbra @Irio I am a developer that worked on that project, maybe I can help with the contact of teacher that idealize this project.

The data that actually have there is true data, but was only added there to some usability tests. The intention is for this year the schoolchildren's will start to add more data there.

Feel free to contact me anytime

caduvieira:
Refers to https://github.com/datasciencebr/serenata-de-amor/issues/42#issuecomment-246142151

cabral:
Teacher reached. Scheduling phase.

Irio:
Any updates on this, @cabral?

cuducos:
Follow up:

@cabral just told me that the he recommended scrapping http://www.consultasocio.com

wisner23:
I'm trying to work on it

cuducos:
Now our source to get Federal Revenue data [offers this data](https://github.com/datasciencebr/serenata-de-amor/issues/16#issuecomment-286167555), so I'm closing this issue ‚ it's a matter of adapting one of our existing scripts and discussing it on #16

PR #115, opened 2016-11-08
Fetch social contract

wisner23:
Following the problem proposed in the #21 . The solution that has been found for this, is to use the site consultasocio.com looking for the name of deputy. But a problem that needs to be solved is in relation to the false positive, because it may happen the name of the deputy searched, not really can be the deputy on the site consultasocio.com. We need to find a way to make the confirmation. The code needs optimization, because actually it takes a lot of time to requests, this can be resolved with some threads.

I'm keep working on it.


src/fetch_social_contract.py
```diff
@@ -0,0 +1,122 @@
...
+
+import requests
+from bs4 import BeautifulSoup
+
+
+BASE_DIR = os.path.dirname(os.path.dirname(os.path.abspath(__file__)))
+DATA_DIR = os.path.join(BASE_DIR, 'data')
+
+
+def download(content_names):
```
cuducos: Why not `deputies` (so you can get rid of line 21)?

Irio:
@cuducos The list of real names is already available at the [Chamber of Deputies' website](http://www2.camara.leg.br/deputados/pesquisa). Here's a direct link (the column is called "Nome Civil"): http://www2.camara.leg.br/deputados/pesquisa/arquivos/arquivo-formato-excel-com-informacoes-dos-deputados-1

cuducos:
@irio This document has exactly 513 congresspeople — probably the ones in charge. It missed the one that's been substituted and the former term names (we have data from the former term and from the former of the former I guess).

PR #117, opened 2016-11-08
A search script to find suspicious places around a geolocalization [Suspect place around searcher]

guilhermeprokisch:
Following the problem proposed in the #39 . In the lasts days, I was thinking how to solve this. I start to trying to create some web crawler for specialized sites of prostitutes, motels, sex shop and etc.. But a problem arise in this approach. Is hard to find specialized websites that contain huge information for all cities in Brazil. But doing this I noticed that almost all this places are in Google Maps with more soft descriptions. Like : "Night Clubs" or "Club of Adult Entertainment". I collected some of this keyword empirically and now I am creating a module to search on the Google Places API for places that match a list of keywords given a geolocalization and return the closest place suspect. I also create a script that uses this module and a similar iterator like _geocode_addresses.py_ to run over all companies dataset and search for the closest suspect place.

- The module are not finished yet and I opened this PR in case someone wants to take a look at the code.

One problem with the Google Places API that I need to point it:
- The free requests per day are 150.000 but each keyword uses one request. Now I'm using 12 keywords for each CNPJ.
So the actual scenario takes more that 5 days to iterate over all companies dataset (I don't know a solution yet)

_I was created another pull request but that was a mess, so I will do it again._


cuducos:
Great contribution, thanks for that!

I was wondering… I see three different approaches to this script:

1. Expanding this script to actually read the suppliers dataset (with latitude and longitude) and generating another dataset with the distances. It might be good because this info would be accessible without tons of HTTP requests in the future. Also we could use the paid API once and save the results.
2. A notebook telling a story about expenses in that kind of place. This is actually what the project is for, this is basically what we promised to deliver.
3. A `src/utils.py` method so people could use this in their own notebook for any kind of analysis.

All these possibilities are non-exclusive and they don't have to be coded by you. Just listing possibilities here.

What do you think about them so we can discuss the right point to merge your contribution?

cuducos:
It looks like you accidentally deleted `config.ini.sample` (probably when renaming it to add your own data). Would you mind re-adding it?

guilhermeprokisch:
Hey Guys..  @cuducos 

I think is finished now. I tested with a small fraction of de companies dataset and works fine.
Add 5 news columns to the dataset 

-  _suspect_place_around_
- _distance_to_suspect 
- _suspect_latitude__ 
- _suspect_longitude_ 
- _suspect_phone_

guilhermeprokisch:
Let me know when you finish run in paid API.  I want see the results. haha

cuducos:
Once more, many thanks for this PR, @guilhermeprokisch! 

@anaschwendler and I worked hard on it today refactoring some pieces, doing some code styling and putting all the pieces to work nicely together. We still haven't got access to the paid API, but this is the only thing we're waiting for merging it — and that should happen pretty soon (@irio, the guy with the API key, was traveling today, so we couldn't use it).

If you wanna follow our commits, pull them from [`suspect-place-searcher` branch](https://github.com/datasciencebr/serenata-de-amor/tree/suspect-place-searcher) — or give us an extra day and I guess everything will be on `master` ; )

src/suspectplacesearcher/main.py
```diff
@@ -0,0 +1,131 @@
...
+                - keyword : The Keyword that generate the place
+                            in Google Place Search
+    '''
+
+    def __init__(self, key):
+        self.GOOGLE_API_KEY = key
+
+    def search(self, lat, lng):
+
+        suspect_keywords = ["Acompanhantes", "Motel",
```
Irio: From where these keywords come from? Are these keywords you manually collected based on existing examples or a pre-existing list?

CONTRIBUTING.md
```diff
@@ -92,8 +92,10 @@ Here we explain what each script from `src/` does for you:
 1. `src/clean_cnpj_info_dataset.py` clean up and translate the supplier info dataset.
 1. `src/geocode_addresses.py` iterate over the supplier info dataset and add geolocation data to it (it uses the Google Maps API set in `config.ini`).
 
+
 ##### Miscellaneous
 1. `src/backup_data.py` uploads files from `data/` to an Amazon S3 bucket set on `config.ini` .
+1. `src/search_suspect_places.py` iterate over the supplier info dataset and add the closest suspect place data to it (it uses the Google Maps API set in `config.ini`). (A "suspect place" is a suspect company nearby: Motels, Massage Houses, Nigh Clubs, Adult Entertainment Houses, etc..)
```
Irio: "Nigh Clubs" -> "Night Clubs"

src/suspectplacesearcher/main.py
```diff
@@ -0,0 +1,131 @@
+# -*- coding: utf-8 -*-
+from geopy.distance import vincenty
+import requests
+import math
+
+
+class SuspectPlaceSearch(object):
```
Irio: "suspect" does not tell much about the meaning we currently have. Is a place selling drugs? A front organization? I suggest something more direct such as "SexRelatedPlace".

src/suspectplacesearcher/main.py
```diff
@@ -0,0 +1,131 @@
...
+                - phone : The phone of suspect place
+                - id : The Google Place ID of suspect place
+                - keyword : The Keyword that generate the place
+                            in Google Place Search
+    '''
+
+    def __init__(self, key):
+        self.GOOGLE_API_KEY = key
+
+    def search(self, lat, lng):
```
Irio: How about `search_nearest` or simply `nearest`, since "search" is already in the class name?

src/search_suspect_places.py
```diff
@@ -0,0 +1,109 @@
...
+is_not_searched_for_suspects = ~data['cnpj'].str.replace(
+    CNPJ_REGEX, '').isin(searched_for_suspects_cnpjs)
+
+remaining_companies = data[is_not_searched_for_suspects]
+
+print('%i companies, %i to go' % (len(data), len(remaining_companies)))
+
+search_suspect_around_companies(remaining_companies)
+
+data = pd.concat([data,
```
Irio: Changing the content of `data/companies.xz` would possibly break existing analyses. Can you create another dataset just with this new information you're bringing to the project? A comma-separated file like the following would be excellent:

```
cnpj,company_name,distance,phone,latitude,longitude
```

This way, those using this dataset can choose the version of the dataset (specially useful when someone updates this script in the future) and merge it with `companies.xz` with a `pandas.DataFrame.merge`.

config.ini.example
```diff
@@ -1,8 +1,9 @@
-[Amazon]
```
Irio: Was this change intended? Not sure if this will work as before.

src/suspectplacesearcher/main.py
cuducos: @guilhermeprokisch's [said](https://github.com/datasciencebr/serenata-de-amor/pull/117#issue-187986747) he had _collected some of this keyword empirically_ — I guess he was listing them manually, trying them out and using the one that made sense according to theses manual tests.

src/search_suspect_places.py
cuducos: I've already fixed that in [`suspect-place-searcher` branch](https://github.com/datasciencebr/serenata-de-amor/tree/suspect-place-searcher), don't worry (or, even better, worry and code review that branch too lol).

src/suspectplacesearcher/main.py
guilhermeprokisch: I don't prefer adopted this name cause SexRelated is too much strict,  being the Class construct to be expansible to search for any kind of company we judge to be Suspect being sexual or not.  Just incrementing the Keyword vector.

src/suspectplacesearcher/main.py
guilhermeprokisch: Once again. I begin to try to construct a crawler to specialized sites on this kind of place (motel, prostitutes, sex massage, etc). But the problem is these websites no contains so much information for all Brazilians regions. But when I start I notice that these places are in google maps with soft descriptions. For example, "Bahamas" (A famous São Paulo Brothel)  are in Google Maps with the description "Adult Entertainment Club"  and this keyword is used for another's brothels in brazil.

PR #122, opened 2016-11-11
Review recommended best practices in the repo

Irio:


cuducos:
I believe it makes a lot of sense. If so, we're changing all notebooks name in this PR, right?

PR #127, opened 2016-11-11
Skip config.ini step in Docker instructions

### LINKED ISSUE #124: Update CONTRIBUTING.md - Docker installation section
In the **Docker Installation Environment** section, topic 2, you are asked to create a _config.ini_ file, with keys to AWS and Google. However, its function is not explained, making it appear that file creation is mandatory. We should refer to the section **Source files (src /)**, subsections **Suppliers information (CNPJ)** and **Miscellaneous**, where the use of the keys is explained.

cuducos:
Fix #124 (and make Docker instructions more coherent with non-Docker instructions).

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -1,15 +1,14 @@
 FROM jupyter/datascience-notebook:latest
-MAINTAINER Serenata de Amor "datasciencebr@gmail.com"
 
 USER root
 
-RUN apt-get update && apt-get install -y \
-  unzip
+RUN apt-get update && apt-get install -y unzip
 
 USER jovyan
 
 COPY requirements.txt ./
 COPY conda_requirements.txt ./
+COPY config.ini.example ./config.ini
 
 RUN pip install --upgrade pip
 RUN pip install -r requirements.txt
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -63,7 +63,6 @@ $ source /usr/local/var/pyenv/versions/anaconda3-4.1.1/bin/activate serenata_de_
 You can user [Docker](https://docs.docker.com/engine/installation/) and [Docker Compose](https://docs.docker.com/compose/install/) to have a working environment:
 
 1. Start the environment (it might take a while, the docker image has 4GB): `$ docker-compose up -d`
-1. Create your `config.ini` file from the example: `$ cp config.ini.example config.ini`
 1. Run the script to download data and other useful files: `$ docker-compose run --rm jupyter python src/fetch_datasets.py`
 1. You can start Jupyter Notebooks and access them at [localhost:8888](http://localhost:8888): `$ docker-compose run --rm jupyter jupyter notebook`
 


Irio:
Why not just running the `setup` executable?

cuducos:
Any specific reason, @dsakuma @gwmoura?

dsakuma:
I think it is possible. But the `setup` run some commands that are not necessary in Docker. I will try to write a configuration that works both in Docker and non-Docker environment.

docker/Dockerfile
```diff
@@ -1,15 +1,14 @@
...
 
-RUN apt-get update && apt-get install -y \
-  unzip
+RUN apt-get update && apt-get install -y unzip
 
 USER jovyan
 
 COPY requirements.txt ./
 COPY conda_requirements.txt ./
+COPY config.ini.example ./config.ini
```
fgrehm: Excuse me for jumping in out of nowhere but this doesn't matter when it comes to local development environments since the whole dir will be overwritten with the project code because of [this](https://github.com/datasciencebr/serenata-de-amor/blob/0a338f09406834c372786984d1a3fe5d6e0383f8/docker-compose.yml#L11). HTH 😄

cuducos: The thing is (I guess) that either user have to manually copy `config.ini.example` or we set the `Dockerfile` to do it — because sooner or later a Python script will look for this file.

What do you think about it? Is there a _best practice_ within Docker community related t environment variables?

fgrehm: Hum... have you guys discussed about just setting them up as `ENV`s on the `Dockerfile`?  (obviously, don't add any secret key there 😅) That way, people just willing to try things out would be a `git clone ... && docker-compose up -d` from getting the env up and running while allowing env vars to be overwritten by adding some entries under the `environment` config of `docker-compose.yml` or just editing the `config.ini`.

But I believe we'll need to change the code that makes use of those settings to prioritize env vars and fallback to the config file right?

cuducos:
Thank you for that. IMHO this is discussion (basically ansent here) is what
is missing (not only in this repo, but in Jarbas too).

I'm a newbie in Docker, so I'll be that annoying person who shares a
concern with no clue about how to fix it: the important thing is to avoid
duplicating code/settings and, at the same time, alow users to get started
using the "normal" (local?) installation process or Docker.

What I mean is that if we set envvars at Dockerfile/docker-compose.yml they
either will be missing in the non-Docker instalation, or would be
duplicated in a INI file. (Again: I'm a noob in Docker and these claims
probably are nonsense. I was just exemplifying my concern here.)

Anyway, from your suggestion we have to inspect how the module that loads
the INI works, am I getting it right? Maybe I can help with that ; )

fgrehm:
> Anyway, from your suggestion we have to inspect how the module that loads
the INI works, am I getting it right? Maybe I can help with that ; )

My suggestion would be to have some sort of abstraction that is able to read configs from env vars and fallback to reading from the ini file. Another idea would be to drop the ini support and just go with using only env vars everywhere.

Last but not least, we could have a custom `ENTRYPOINT` that sets things up (ex: copy the `.ini` file to the right place) before running a given command specified either on the `Dockerfile` itself or provided by `docker run` / `docker-compose.yml`

cuducos:
> My suggestion would be to have some sort of abstraction that is able to read configs from env vars and fallback to reading from the ini file. Another idea would be to drop the ini support and just go with using only env vars everywhere.

I guess [`python-decouple`](https://pypi.python.org/pypi/python-decouple) can handle that — what do you think about it. There are more robust solutions, such as [`dynaconf`](https://github.com/rochacbruno/dynaconf) too.

Issue #102, opened 2016-11-03
Collecting restaurant menu prices


anaschwendler:
One of the hypothesis that we are planning to work on the next two months is about collecting and evaluating restaurant menu prices. We don't know where is the best place to collect these data, but we're thinking about collecting from Google Maps, Yelp or Foursquare.

Ideally we will use those data to return the probability of the hypothesis that a given expense is a suspcious outlier in its group.

anaschwendler:
News: 
Restaurant prices changes everyday and we cannot count with the prices by now for a 2011 receipts, for example. So, talking with @cuducos and @Irio i was thinking in better ways to get the prices, and what passed in my mind is that we can use the data we already have from the receipts from CEAP to get an average cost for restaurants (thinking that there are more than one receipt for some restaurants).
By doing that we can already find some outliers, it's not the best way, but it's a start.

@cuducos suggested that we can use foursquare and google maps to check how expensive a restaurant is (by the rating that they have) and use that to match with the receipts we have from the restaurants and if it fits to what we got.

thcavalcante:
Trip advisor is a good resource to find out the average too.

talespaiva:
Hi, I took some advice from this thread to create the PR #129. Hope it's not duplicated work.

rafaelcostadefarias:
Some prices can be collected at ifood. Maybe with a webcrawler?

cuducos:
Hi @rafaelcostadefarias,

> Some prices can be collected at ifood. Maybe with a webcrawler?

IMHO this is not the best approach due to what @anaschwendler [has explained above](https://github.com/datasciencebr/serenata-de-amor/issues/102#issuecomment-259171939) — but thanks for the idea anyway ; )

filipelinhares:
Hi guys. I'm working on a script to collect info from [Yelp](https://yelp.com).

Irio:
@cuducos I believe @rafaelcostadefarias' idea is good even with @anaschwendler's proposal. The repo may accept multiple analysis in the same subject.

cuducos:
@Irio: not sure if we're talking about new analysis (in my mind: new `.ipynb`) or new datasets (in my mind: new CSVs).

[...]

PR #129, opened 2016-11-12
Meal analysis

talespaiva:
One of the ideas of #102 is to use information already on the data as a first step on identifying unusual meal expenses. This PR, contains a Notebook that uses a robust oulier method to indentify some suspect cases.


anaschwendler:
Hi @talespaiva first of all, thank you so much for your contribuition, it's not a duplicate work, and it helps a lot. You did a great job at all! :)
I was looking at your code and notice some things, can you explain a little bit more of what you are trying to do in your code and the next-steps?

anaschwendler:
I think that the best way to continue this work is generating a dataset with the average prices that we found for all places, that is something that @talespaiva explored, what do you think about that?

talespaiva:
Great. Will try to better explain the notebook. I added some comments in between, but apparently it didn't help a lot. In fact, re-reading the notebook now made me realize that it's a little mixed up.

The whole analysis is based on grouping only meal expenses by CNPJ/CPF. Up to cell [8], it's just data preparation. First, I've looked to the places that received most money (cell [9]). The first issue with the dataset is that some places have more than one CNPJ, like SENAC. The accumulated net_value does not say much about irregularities, so I've reordered the meals dataset to reveal the most visited places (cell [11]). From there, I combined the two previous dataframes to have an average value per visit (cell [13]), that has shown some strange values such as R$ 4200 spent on just one visits in a place that looks like a car selling store (according to its name). I have even contacted Lucio Big from OPS about this case and he said this congressperson had already been sued by him for things like that.

From this first part, I think that a future work could be identifying those places that are almost never visited and have high values spent on. There's not enough data to take an average value, but an overall price for the entire dataset may be useful to highlight these cases.

At cell [15], I've made a rather naïve assumption that the more a place is visited, the more money is spent in there. This is obviously not true since there are great differences in prices depending on which place is visited, but it turns out that a linear model wasn't too bad, but I think other models should be tested and compared later. From this part, the points that are too far *below* the line in the picture could be better investigated because they are either too expensive (they get more money with less visits).

From cell [16] on, I tried to make an individual analysis, more in the spirit of #102, by using only expenses in a single place (notice that there the two SENAC instances should be merged for a proper anaylis, but it wasn't done in this case). For that, I've used the [Outlier Labeling Rule](http://www.itl.nist.gov/div898/handbook/eda/section3/eda35h.htm), that uses the median instead of the mean, a measure that is more robust to outliers. The authors of this methods recommends that any modified z-score greater than 3.5 should be considered as a potential outlier if the data is expected to come from a normal distribution. Basically, what follows is the application of this method for the first 10 most visited places, as it's not very helpful to make this to places less frequented.

I like the idea of creating a dataset with the average prices (may be with the **median** price too), however I think the number of visits should be included. Also, merging CNPJs must be addressed in this task (I don't know if it's only the case of SENAC) and I found that selecting a name to associate to a CNPJ/CPF let the analysis more pleasant to the eyes :-)

Another branch that can be created for the project is identifying names of places that don't have much to do with the subquota under it was paid, for example, the case of meal in a car selling store. I don't know if this is on your roadmap, just a random idea here.

I think I repeated some things already explained in the notebook. If it wasn't clear, please let me know if I can better explain some point.

anaschwendler:
@talespaiva that's awesome, thanks for the explanation, but don't you think that those comments about you code would fit better in your code? :)

> I like the idea of creating a dataset with the average prices (may be with the median price too), however I think the number of visits should be included. Also, merging CNPJs must be addressed in this task (I don't know if it's only the case of SENAC) and I found that selecting a name to associate to a CNPJ/CPF let the analysis more pleasant to the eyes :-)

Couldn't agree more, I think that's a great idea, it will be nice to have a dataset that looks like this.

> Another branch that can be created for the project is identifying names of places that don't have much to do with the subquota under it was paid, for example, the case of meal in a car selling store. I don't know if this is on your roadmap, just a random idea here.

Yes, it's in the roadmap, but it can be better investigated using machine learning algorithms

> I think I repeated some things already explained in the notebook. If it wasn't clear, please let me know if I can better explain some point.

No, the notebook it's awesome, I was talking to @cuducos and I really wanted to do something like that. That's a great contribuition, I'm only thinking about doing a dataset, or what can we do more with this first step, for us to explore this data in a better way!

talespaiva:
@anaschwendler, I have updated the notebook with some parts of my comment. I can help with creating the dataset, but I'm busy for the next few days.

anaschwendler:
@talespaiva that was exatcly what i was looking for, i will work based on what you had collected and try to create a dataset.

Thanks a lot for the contribuition, it saved a lot of time for us :heart:
Soon, i'll merge your pr to the master branch! :)

anaschwendler:
It's up to go :)

Merging it!

Issue #120, opened 2016-11-10
Collecting civil names of congressperson


anaschwendler:
In a talk with @wisner23 and @cuducos we discovered that we only have the congressperson name, not their civil names. Getting those names is relevant and important to our project, since we can by using those names, explore more datasets in a better way.

The idea is to create a dataset that associate civil name to congressperson name, using the data from [câmara](http://www2.camara.leg.br/deputados/liderancas-e-bancadas/bancadas/bancada-atual). Remembering that our congressperson_id matches with the id of congressperson of câmara website :)

wisner23:
i'm try work on it

Irio:
http://www2.camara.leg.br/transparencia/recursos-humanos/quadro-remuneratorio/consulta-deputados

anaschwendler:
@Irio That link has only the actual congressperson elected. With that script we fetch all congressperson civil names, including from other elections.

anaschwendler:
fix #120

cuducos:
Fixed by #132

anaschwendler:
@wisner23 thanks for you work again, it's on! :tada:

wisner23:
I always I'll be helping ;P

PR #132, opened 2016-11-13
Fetch civil names

wisner23:
Following the issue #120 .The solution found was to do a scraping on http://www.camara.leg.br.


src/fetch_civil_names.py
```diff
@@ -0,0 +1,101 @@
+import datetime
+import os
+import requests
+import re
+import pandas as pd
+import numpy as np
+from bs4 import BeautifulSoup
+
+
+DATA_BASE_DIR = os.path.dirname(os.path.dirname(os.path.abspath(__file__)))
```
cuducos: Actually this is the base dir of the project, isn't it? Should we call it `BASE_DIR`?

src/fetch_civil_names.py
```diff
@@ -0,0 +1,101 @@
...
+    newest_file = find_newest_file(name)
+    if newest_file is None:
+        raise TypeError('Could not find the dataset for {%s}.' % newest_file)
+
+    return pd.read_csv(newest_file, dtype={
+                                        'congressperson_id': np.str
+                                    })
+
+
+def get_all_congress_people_ids():
```
cuducos: 💅 `get_all_congresspeople_ids` — congresspeople is one word ; )

src/fetch_civil_names.py
```diff
@@ -0,0 +1,101 @@
...
+    print('Fetching all congress people ids...')
```
cuducos: 💅 idem: _congresspeople_ is one word ; )

src/fetch_civil_names.py
```diff
@@ -0,0 +1,101 @@
...
+
+    ids = (read_csv(name)['congressperson_id'] for name in ['current-year', 'last-year', 'previous-years'])
+    return list(str(idx).replace('\n', '').strip() for idx in pd.concat(ids).unique())
```
cuducos: In fact `.strip()` already cleans what `.replace('\n', '') is cleaning:

```python
>>> s = """
...    42
...
...
... """
>>> s
'\n   42    \n\n\n'
>>> s.strip()
'42'
>>>
```

And resource wise a generator would work better than `return list(…)`.

I'd recommend changing this line for:

```python
yield from (str(idx).strip() for idx in pd.concat(ids).unique())
```

(You can also try `idx.strip()` — as `congressperson_id` is converted to `np.str` it might have the `.strip()` method but I don't remember by heart.)

src/fetch_civil_names.py
```diff
@@ -0,0 +1,101 @@
...
+    df = pd.DataFrame(data=congressperson_civil_names)
+    df.to_csv(os.path.join(DATA_PATH, FILE_BASE_NAME), **CSV_PARAMS)
+
+    print('Done.')
+
+
+def get_civil_names():
+
+    congress_people_ids = get_all_congress_people_ids()
+    total = len(congress_people_ids)
```
cuducos: **If** get_all_congress_people_ids() is a generator, `len(generator)` will not work. Maybe we can get the total by other means.

src/fetch_civil_names.py
```diff
@@ -0,0 +1,101 @@
...
+    congress_civil_name = []
+
+    for ind, congress_id in enumerate(congress_people_ids):
+
+        page = requests.get(URL_SCRAPING.format(congress_id))
+        data = str(page.content.decode('utf-8'))
+
+        if page.status_code != 200:
+            raise TypeError('could not request')
```
cuducos: You can print the current `status_code`: it might help in knowing why the request failed (were we blocked? malformed URL?). I'd print `'HTTP request to {} failed with status code {}'.format(url, page.status_code)` or something like that.

src/fetch_civil_names.py
```diff
@@ -0,0 +1,101 @@
...
+    df.to_csv(os.path.join(DATA_PATH, FILE_BASE_NAME), **CSV_PARAMS)
+
+    print('Done.')
+
+
+def get_civil_names():
+
+    congress_people_ids = get_all_congress_people_ids()
+    total = len(congress_people_ids)
+    congress_civil_name = []
```
cuducos: Instead of using a blank list with `.append(…)` I would recommend converting this function into a generator. Just `yield dict(congressperson_id=congress_id, civil_name=line_name.text.strip())` by line 89 and we're good to go.

wisner23:
Good tips @cuducos ! i already performed the fixes, sorry for the trouble on commits.

cuducos:
> sorry for the trouble on commits

I haven't event noticed them, man ; )

anaschwendler:
Hi @wisner23 I was taking a closer look at your results, great job!
But going a little further i started looking the name you couldn't get, and i guess that they are the congressperson from another elections.

For example, the first id that your script couldn't find was: #74037 and you can find the result [here](http://www2.camara.leg.br/deputados/pesquisa/layouts_deputados_biografia?pk=74037). What do you think about considering fetching from here when you don't find in the first link?

wisner23:
i' ve implemented the new repository to fetch, using two repositories, we got found 1147 civil names ! Thanks for you tip @anaschwendler!

anaschwendler:
Hi @wisner23!
First of all: great job! Thank you so much for your contribuition!

Me and @cuducos made some simple refactoring in your code and merged it, now it's on :)
By now, thanks for the work, looking forward to work with you again o/

Any doubts, we are here to help!

anaschwendler:
Fix #120

anaschwendler:
Fix #120

anaschwendler:
fixes #120

anaschwendler:
fix #120

cuducos:
Fix #120

Issue #135, opened 2016-11-15
net_value limits


pmargreff:
Reading [the glossary](http://www.camara.gov.br/cota-parlamentar/2016_Glossario_para_o_CotasNet.pdf) I found some limits to the `subquota_description` so, I tried to find moments this value is bitten.  

The first analysis the max was bitten in R$ 11 M, so I look again and saw the dates the values start to be valid. 

In the second moment following the date the values are validates we can recovered R$ 500 k. I think have made some mistake, I don't believe the value is so large anyway. But, it could get some irregularities, and maybe after the net_value be fixed it could get a few more cases. 

Here is all dates with reference on code: 

[ATO DA MESA Nº 112, DE 15/10/2013](http://www2.camara.leg.br/legin/int/atomes/2013/atodamesa-112-15-outubro-2013-777266-publicacaooriginal-141483-cd-mesa.html)
[ATO DA MESA Nº 4, DE 25/02/2015](http://www2.camara.leg.br/legin/int/atomes/2015/atodamesa-4-25-fevereiro-2015-780188-publicacaooriginal-146197-cd-mesa.html) 
[ATO DA MESA Nº 49 DE 03/09/2015](http://www2.camara.leg.br/legin/int/atomes/2015/atodamesa-49-3-setembro-2015-781493-publicacaooriginal-148023-cd-mesa.html)
[ATO DA MESA Nº 142, DE 28/05/2016](http://imagem.camara.gov.br/Imagem/d/pdf/DCD0020140529SA0790000.PDF#page=3)

pmargreff:
My idea is compare campaign date with the expenses date in publicity, and rewrite and try validate if the `subquota_description` have some coherence with the product/service .

anaschwendler:
Hi @pmargreff!
So, we were starting to apply those rules in a script yesterday, but we find some 
inconsistencies in out dataset (that can be found in: https://github.com/datasciencebr/serenata-de-amor/issues/133) while we were developing it.

By now, we are waiting for a reply from câmara, I think that is better to analyse this before we have the right dataset, what do you think?

pmargreff:
@anaschwendler, ok, thanks for spent some time in. I looked into the pull request made and didn't find anything related with (or I couldn't understand). Anyway, I'll wait the reply to try again.

cuducos:
>  I looked into the pull request made and didn't find anything related with

TL;DR:

* we're not sure if we can take into account rows without `reimbursement_number`.
[...]

PR #136, opened 2016-11-15
feature: checking if the net_value pass the limit in subquota with limit defined

pmargreff:
Commit related with issue #135.


pmargreff:
The values get over the limit aren't so high how I imagined, anyway I would know the procedure to demand the values back from the deputies? 

Another nice nuance i get in some moment, for the fuel spent have exactly the same value for a lot of people, it's almost impossible, so in a near future, I'll check the values to see if they came from the same source, and how it's distributed through the fuel companies. It's very suspicious.

pmargreff:
Any reason for not accept that? Did you try the refund from these values?

cuducos:
Hi @pmargreff — sorry about taking so long to get back, and thanks for the contribution (and the reminder).

I'd say that before merging that we have to address the issue about making Julia available in the repo setup, as @anaschwendler [mentioned the other day](https://github.com/datasciencebr/serenata-de-amor/issues/135#issuecomment-262855451).

cuducos:
Closed by inactivity — but anyone wiling to bring Julia in is welcomed to re-open this and allow everyone to run Julia on notebooks easily ; )

PR #144, opened 2016-11-21
Meal prices scrapping

cabral:
This PR is a pre-scrapping for meal scrapping and price comparison. 
Each congressperson is allowed to buy food only for himself.

For websites like Tripadivsor and Foursquare that use $, we translate them to integer and add a column to the dataset for comparison with the meal bill. 
Also I've added the meal filter to the whole dataset and cleaned the CNPJ(enterprise identity number) - creating a dataset that is possible to be used with api's

### DIFF
--- a/develop/2016-11-16-cabral-meal-price-scrapping.html
+++ b/develop/2016-11-16-cabral-meal-price-scrapping.html
@@ -1,0 +1,12697 @@
+<!DOCTYPE html>
+<html>
+<head><meta charset="utf-8" />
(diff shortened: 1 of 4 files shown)


cuducos:
@cabral just to confirm: changes in `utils.py` are yours or is this some git issue? Also can you check this conflict in `src/fetch_civil_names.py`? (We can pair and sor that out quickly I guess).

cabral:
Already served his purpose - have been integrated in yelp and ifood scrapping scripts by @filipelinhares

PR #146, opened 2016-11-23
Script to fetch foursquare data about companies in meal reimbursements

andrepinho:
It tries to find a match in foursquare API using our companies dataset. It uses trade_name, zip_code, latitude and longitude values.

### DIFF
--- a/src/fetch_foursquare_info.py
+++ b/src/fetch_foursquare_info.py
@@ -52,6 +52,7 @@ def load_companies_dataset(cnpjs):
                                 low_memory=False,
                                 usecols=u_cols,
                                 dtype={'trade_name': np.str})
+    all_companies = all_companies.dropna(subset=['cnpj', 'trade_name'])
     all_companies['clean_cnpj'] = all_companies['cnpj'].map(only_numbers)
     return all_companies[all_companies['clean_cnpj'].isin(cnpjs)]
 
@@ -83,22 +84,29 @@ def search(company):
 
     params.update({'query': company['trade_name'],
                    'll': '%s,%s' % (company['latitude'], company['longitude']),
-                   'zip_code': company['zip_code']})
-
-    if CONFIRMED_MATCHES_ONLY:
-        params.update({'intent': 'match'})
+                   'zip_code': company['zip_code'],
+                   'intent': 'match'})
 
     url = 'https://api.foursquare.com/v2/venues/search'
     response = requests.get(url, params=params)
-    return parse_search_results(response)
+    result = parse_search_results(response, True)
+
+    if not result:
+        params.pop('intent')
+        response = requests.get(url, params=params)
+        result = parse_search_results(response, False)
+
+    return result
 
 
-def parse_search_results(response):
+def parse_search_results(response, confirmed_match):
     """Return the first venue from the given search response"""
     json_response = response.json()
     venues = json_response.get('response', {}).get('venues')
     if venues:
-        return venues[0]
+        venue = venues[0]
+        venue['confirmed_match'] = confirmed_match
+        return venue
 
 
 def fetch_venue(venue_id):
@@ -134,12 +142,6 @@ DEFAULT_PARAMS = {'client_id': CLIENT_ID,
                   'client_secret': CLIENT_SECRET,
                   'v': VERSION}
 
-# This variables defines whether the search method is going to run with the
-# param "intent=match" or not. This is also save as "confirmed_match" column
-# for every record being fetched.
-# More info about it: https://developer.foursquare.com/docs/venues/search
-CONFIRMED_MATCHES_ONLY = True
-
 # Dataset paths
 REIMBURSEMENTS_DATASET_PATH = find_newest_file('reimbursements')
 COMPANIES_DATASET_PATH = find_newest_file('companies')
@@ -165,7 +167,6 @@ if __name__ == '__main__':
             fetched['cnpj'] = company['cnpj']
             fetched['clean_cnpj'] = company['clean_cnpj']
             fetched['scraped_at'] = datetime.datetime.utcnow().isoformat()
-            fetched['confirmed_match'] = CONFIRMED_MATCHES_ONLY
             fetched = json_normalize(fetched)
 
             fetched_companies = pd.concat([fetched_companies, fetched])
--- a/src/fetch_foursquare_info.py
+++ b/src/fetch_foursquare_info.py
@@ -8,24 +8,22 @@ import re
 import requests
 from pandas.io.json import json_normalize
 
-# Dataset paths
-REIMBURSEMENTS_DATASET_PATH = os.path.join('data', '2016-11-22-reimbursements.xz')
-COMPANIES_DATASET_PATH = os.path.join('data', '2016-09-03-companies.xz')
-FOURSQUARE_DATASET_PATH = os.path.join('data', 'foursquare-companies.xz')
+def find_newest_file(name):
+    """
+    Assuming that the files will be in the form of :
+    yyyy-mm-dd-type_of_file.xz we can try to find the newest file
+    based on the date, but if the file doesn't exist fallback to another
+    date until all dates are exhausted
+    """
+    date_regex = re.compile('\d{4}-\d{2}-\d{2}')
 
-# API Keys
-# You can create your own through https://pt.foursquare.com/developers/register
-settings = configparser.RawConfigParser()
-settings.read('config.ini')
-CLIENT_ID = settings.get('Foursquare', 'ClientId')
-CLIENT_SECRET = settings.get('Foursquare', 'ClientSecret')
-#Foursquare API Version. This is in YYYYMMDD format.
-VERSION = '20161021'
-
-# Required params to make a request to Foursquare's API
-DEFAULT_PARAMS = {'client_id': CLIENT_ID,
-                'client_secret': CLIENT_SECRET,
-                'v': VERSION}
+    matches = (date_regex.findall(f) for f in os.listdir(DATA_DIR))
+    dates = sorted(set([l[0] for l in matches if l]), reverse=True)
+    for date in dates:
+        filename = os.path.join(DATA_DIR, '{}-{}.xz'.format(date, name))
+        if os.path.isfile(filename):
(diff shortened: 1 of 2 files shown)


src/fetch_foursquare_info.py
```diff
@@ -0,0 +1,129 @@
...
+import datetime
+
+# Setting dataset paths
+REIMBURSEMENTS_DATASET_PATH = os.path.join('data', '2016-11-22-reimbursements.xz')
+COMPANIES_DATASET_PATH = os.path.join('data', '2016-09-03-companies.xz')
+FOURSQUARE_DATASET_PATH = os.path.join('data', 'foursquare-companies.json')
+
+# These are the keys to access Foursquare's API. You can create your own through the link bellow
+# https://pt.foursquare.com/developers/register
+CLIENT_ID = ''
```
cuducos: I'd suggest using `config.ini` as in [`src/fetch_datasets.py`](https://github.com/datasciencebr/serenata-de-amor/blob/master/src/fetch_datasets.py#L23-L26) to keep API keys safe and restrict to environment variables (not to the repo).

src/fetch_foursquare_info.py
```diff
@@ -0,0 +1,129 @@
...
+# Required params to make a request to Foursquare's API
+DEFAULT_PARAMS = {'client_id': CLIENT_ID,
+                'client_secret': CLIENT_SECRET,
+                'v': VERSION}
+
+def load_cnpjs(subquota_description):
+    u_cols = ['cnpj_cpf', 'subquota_description']
+    docs = pd.read_csv(REIMBURSEMENTS_DATASET_PATH,
+                        low_memory=False,
+                        usecols=u_cols,
```
cuducos: Kudos for introducing me to the `usecols` argument 💕

src/fetch_foursquare_info.py
```diff
@@ -0,0 +1,129 @@
...
+import numpy as np
+import pandas as pd
+import re
+import json
+import datetime
+
+# Setting dataset paths
+REIMBURSEMENTS_DATASET_PATH = os.path.join('data', '2016-11-22-reimbursements.xz')
+COMPANIES_DATASET_PATH = os.path.join('data', '2016-09-03-companies.xz')
+FOURSQUARE_DATASET_PATH = os.path.join('data', 'foursquare-companies.json')
```
cuducos: Can we save data in the standard format for the project? CSV compressed as `.xz`?

src/fetch_foursquare_info.py
```diff
@@ -0,0 +1,129 @@
+import os.path
+import requests
+import numpy as np
+import pandas as pd
+import re
+import json
+import datetime
+
+# Setting dataset paths
+REIMBURSEMENTS_DATASET_PATH = os.path.join('data', '2016-11-22-reimbursements.xz')
```
cuducos: Maybe we could dynamically load the most recent `YYYY-MM-DD-reimbursements.xz` instead of a fixed date (the same for `COMPANIES_DATASET_PATH`).

src/fetch_foursquare_info.py
```diff
@@ -0,0 +1,129 @@
...
+    fetched_companies = load_foursquare_companies_dataset()
+    return companies[~companies['cnpj'].isin(fetched_companies['cnpj'])]
+
+def load_foursquare_companies_dataset():
+    if os.path.exists(FOURSQUARE_DATASET_PATH):
+        return pd.read_json(FOURSQUARE_DATASET_PATH)
+    else:
+        return pd.DataFrame(columns=['cnpj'])
+
+def get_venue(company):
```
cuducos: What kind of data is `company`? A Pandas row? A CNPJ string? A parsed JSON object? I'd suggest adding [`__docstrings__`](https://www.python.org/dev/peps/pep-0257/) to (almost) all functions to make your logic clearer…

src/fetch_foursquare_info.py
```diff
@@ -0,0 +1,129 @@
...
+    params.update({
+        'query': company['trade_name'],
+        'll': '%s,%s' % (company['latitude'], company['longitude']),
+        'zip_code': company['zip_code'],
+        'intent': 'match'
+    })
+    url = 'https://api.foursquare.com/v2/venues/search'
+    response = requests.get(url, params=params)
+
+    return parse_search_results(response)
```
cuducos: I'd check response's HTTP status here before passing it to the parser (same for line 85). We should rely on the server header, not in the JSON contents to know how successful was the request.

Something like:

```python
def search(company, timeout=5):
    …
    response = requests.get(url, params=params)
    if 200 <= response.status_code < 300:
        return parse_search_results(response)
```

andrepinho:
@cuducos Thanks a lot for your review :) Just pushed a commit taking what you said into consideration. The only thing I've left out is the part where you suggested to fetch the most recent dataset. I remember seing some code to do that but don't remember where. As soon as I find it (do you recall seeing it aswell?), I'll implement in this script.

About performance: With 2 requests per company, my modest laptop (i5 4x1.7GHz) is already faster than what [foursquare's limits](https://developer.foursquare.com/overview/ratelimits)(5k requests/hour) allow. If we want to go faster, we'll need multiple API keys and some more code to handle it. Do you think it's worth the trouble?

src/fetch_foursquare_info.py
```diff
@@ -32,98 +37,99 @@ def load_cnpjs(subquota_description):
...
     url = 'https://api.foursquare.com/v2/venues/search'
     response = requests.get(url, params=params)
-
-    return parse_search_results(response)
+    if 200 >= response.status_code < 300:
+        return parse_search_results(response)
 
 def parse_search_results(response):
+    """Return the first venue from the given search response"""
     json = response.json()
```
cuducos: Be careful: in this scope you're overwriting the native `json` module (imported in line 3). Yep, naming things is the toughest part of coding!

src/fetch_foursquare_info.py
```diff
@@ -32,98 +37,99 @@ def load_cnpjs(subquota_description):
...
-    if json['meta']['code'] == 200:
-        venues = json['response']['venues']
-        if venues:
-            return venues[0]
+    venues = json['response']['venues']
```
cuducos: This might be a bit of 💅 but…

If you get a JSON a bit different than expected `json['response']['venues']` will crash with a `KeyError`. You could use [Dictionary's `get` method](https://docs.python.org/3/library/stdtypes.html?highlight=get#dict.get) to overcome that:

```python
json_response = response.json()
venues = json_response.get('response', {}).get('venues')
```

(In the first `get` I'm setting the `default` to an empty dictionary (`{}`) so if the key `response` is missing the returned object  still have the `get` method (if the first `get` returns `None` the second `get` fails because `None` não tem o método `get`).

src/fetch_foursquare_info.py
```diff
@@ -32,98 +37,99 @@ def load_cnpjs(subquota_description):
...
-def fetch_venue_by_id(venue_id):
+def fetch_venue(venue_id):
+    """Return specific data from Foursquare for the given venue_id"""
     url = 'https://api.foursquare.com/v2/venues/%s' % venue_id
     response = requests.get(url, params=DEFAULT_PARAMS)
-    return parse_venue_info(response)
+    if 200 >= response.status_code < 300:
+        return parse_venue_info(response)
 
 def parse_venue_info(response):
```
cuducos: Taking into account what I just said about `parse_search_results(…)`, this could be simplified as:

```python
def parse_venue_info(response):
    """Return only venue data from the given fetch_venue response"""
    json_response = response.json()  # don't overwrite native json module
    return json_response.get('response', {}).get('venue')  # returns the venue info or None
```

Surely you can delete my comments ; )

cuducos:
*About fetching the most recent dataset*: I've sent you in PVT but as more people might be following this PR. [here it is](https://github.com/datasciencebr/serenata-de-amor/blob/master/src/get_family_names.py#L82-L98) (@irio this tiny function is so useful we might add it to [serenata-toolbox](https://github.com/datasciencebr/serenata-toolbox) — I mean, if it isn't already there).

*About performance*: if you think it's a good pace, I believe it ; ) Can you estimate how log it will take to run all the script through the whole dataset?

Irio:
@cuducos If it's getting used in multiple places, `src/utils.py` is the place to be. :)

andrepinho:
@cuducos Pushed another commit with the changes you suggested.

_About performance_: I'm not sure how to estimate that (ideas?). I know that right now the bottleneck is definitely foursquare's limit. I'm reaching it everytime I run the script for the whole dataset.
Something to improve would be not to look for companies that hasn't been found in a previous run. Otherwise, we're wasting requests with companies we know aren't going to be matched (unless the script is changed or foursquare has recent info about those companies). I'm not sure if and how I should tackle this.

cuducos:
> I'm not sure how to estimate that (ideas?)

Just a rough guess is enough. If the dataset has 60k rows, check how long it takes to look for the first 2k and multiply by 30. If the API limit is 5k/hour, we can use count requests and time and make the script pause if 5k requests are reached before 1 hour… but I think this is _too much_. Just re-reunning and skipping the ones that are already in the dataset might be enough (maybe wrote to the dataset the ones with no replies, just for control).

andrepinho:
Pushed some more adjusts and removed the WIP label. Now it has a configuration variable to change it from "intent=match" to regular search (WAY more matches but higher chance of false positives). The idea is to run with it untill we stop finding results and then switch to the regular search for the remaining companies.

There are around 20k unique CNPJs from "Congressperson meal" reimbursements. I left the script running for about 1h and it went through ~4k companies. It found the last result somewhere between 3.5~3,6k and then foursquare's limit kicked in and it only returned "No results". In this time, it matched 1295 companies with CONFIRMED_MATCHES_ONLY set to True.

andrepinho:
Pushed yet more changes. Did it paired with @anaschwendler. Thanks a lot Ana <3

We skipped companies without "trade_name". We tried to use the "name" column when "trade_name" was unavailable but it prooved to be a bad call when running the script without the "intent=match" param. It brought too many false matches.

Speaking about "intent=match" now the script automatically removes the param and searches again if no result is matched with it. It also handles the extra column "confirmed_match".

Issue #13, opened 2016-08-21
Calculate and find anomalies on traveled speeds from last meal


Irio:
The Quota for Exercise of Parliamentary activity says that meal expenses can be reimbursed just for the politician, excluding guests and assistants. Creating a feature with information of traveled speed from last meal can help us detect anomalies compared to other expenses.

For doing so, we need the location of each expense. One possible way of getting that is fetching the address of each CNPJ receiving money from expenses with "Congressperson meal" as `subquota_description`.

leportella:
I am trying to work on it

anaschwendler:
@leportella by the way, great work! Can you tell us more of what you've done?
I think that the way now is finding the analysis about distances and time differentes.

PR #150, opened 2016-11-30
Analysis: traveled speeds between meals

Irio:
Proposed in https://github.com/datasciencebr/serenata-de-amor/issues/13.

The CEAP says that meal expenses can be reimbursed just for the politician, excluding guests and assistants. Creating a feature with information of "traveled speed" (i.e. too many meals in distant cities, in a short period of time) from last meal can help us detect anomalies compared to other expenses.

Since we don't have in structured data the time of the expense (just in the picture of the receipt, when available), we want to analyze the group of expenses made in the same day/week.

Anomalous days are limited in 0.1% of the dataset, or 125 deputies' days. This gives us 703 reimbursements to investigate, summing in R$ 28752.

---

Will require existing contributors to run `$ conda install --yes --file conda_requirements.txt` after getting merged. For those who don't have an environment set up, `$ setup` will do the job.


anaschwendler:
Nice job, looking for forward to work with you again :)

PR #156, opened 2016-12-08
Docker tweaks

fgrehm:
Hey guys! Great work on this project, 🍻 / 🍷 / ☕️ to everyone involved 😄 

While I'm still trying to find out how can I help on the data side of things, I noticed some improvements we can make to the Docker dev environment, namely reducing the amount of layers and size of the resulting image:

screen shot 2016-12-08 at 12 05 29 pm

_The second image is the one I built with the code that is on master_

... and setting up the automagic `.html` / `.py` creation when saving notebooks.

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -3,18 +3,22 @@ MAINTAINER Serenata de Amor "datasciencebr@gmail.com"
 
 USER root
 
-RUN apt-get update && apt-get install -y \
-  unzip
+RUN apt-get update \
+    && apt-get install unzip -y --no-install-recommends \
+    && apt-get clean \
+    && rm -rf /var/lib/apt/lists/*
 
 USER jovyan
 
 COPY requirements.txt ./
+RUN pip install --upgrade pip \
+    && pip install -r requirements.txt
+
 COPY conda_requirements.txt ./
+RUN conda update --yes conda \
+    && conda config --add channels Rufone \
+    && conda config --add channels conda-forge \
+    && conda install --yes --file conda_requirements.txt \
+    && conda clean -tipsy
 
-RUN pip install --upgrade pip
-RUN pip install -r requirements.txt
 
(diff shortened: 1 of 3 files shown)


docker/Dockerfile
```diff
@@ -3,18 +3,22 @@ MAINTAINER Serenata de Amor "datasciencebr@gmail.com"
...
-RUN apt-get update && apt-get install -y \
-  unzip
+RUN apt-get update \
+    && apt-get install unzip -y --no-install-recommends \
+    && apt-get clean \
+    && rm -rf /var/lib/apt/lists/*
 
 USER jovyan
 
 COPY requirements.txt ./
```
fgrehm: BTW, the idea behind splitting up the copy & installs here is to leverage Docker caching mechanism by making sure we don't reinstall conda packages because of python dependencies changes

docker/Dockerfile
```diff
@@ -3,18 +3,22 @@ MAINTAINER Serenata de Amor "datasciencebr@gmail.com"
...
+    && pip install -r requirements.txt
 
-RUN pip install --upgrade pip
-RUN pip install -r requirements.txt
+COPY conda_requirements.txt ./
+RUN conda update --yes conda \
+    && conda config --add channels Rufone \
+    && conda config --add channels conda-forge \
+    && conda install --yes --file conda_requirements.txt \
+    && conda clean -tipsy
```
vmesel: @fgrehm probably these are default commands on conda install.

fgrehm: Sorry, not sure what you meant =/

Just to make it clear, the idea with this change is to [remove the cache](http://conda.pydata.org/docs/commands/conda-clean.html) from the docker image and reduce the image size. The reason I said I dunno if it has any side effect is that the help says it should remove unused packages but I'm not sure how it detects unused packages since this all new to me 😄

PR #157, opened 2016-12-08
Fix indentation on CEAP doc

fgrehm:
Super minor, but I think this reads better 😄


Irio:
Happy first contribution day! 🎉

PR #161, opened 2016-12-08
Validate cnpj_cpf field looking for fake values

Irio:
`cnpj_cpf` is the column identifying the company or individual who received the payment made by the congressperson. Having this value empty should mean that it's an expense made outside Brazil, with a company (or person) without a Brazilian ID.

10 suspicious reimbursements are returned.

---

This code is almost ready to be implemented in [Rosie](https://github.com/datasciencebr/rosie) as a new "classifier", creating a new irregularity column. Any volunteer to bring this code (+ tests) there?


vmesel:
is this a PR or an issue?

cuducos:
Irio, may I suggest a minor change before merging? You could use `document_type` to get only CNPJ and CPF (discarding the expenses made abroad) and then runing the validator. Does that make sense?

Irio:
@cuducos From [the docs](http://jarbas.datasciencebr.com/static/ceap-datasets.html):

screen shot 2016-12-10 at 9 00 08 pm

In other words, `document_type` stands for something else

cuducos:
So… Doesn't discarding `document_type==2` (expenses made abroad) leave you only with rows either with a CNPJ or a CPF? Anyway… minor detail, nothing really meaningful here, sorry.

Irio:
@vmesel It's a pull request, but may become an issue in Rosie, asking for help to implement it there.

PR #164, opened 2016-12-12
Fixed a few mistakes on CONTRIBUTING.md

marceldalagnol:
I made a few changes to the CONTRIBUTING.md file, in order to fix small mistakes and help with clarity. I tried to keep changes to a minimum, since all I know about the code came from that same file (hardly in-depth). I apologize if any of the changes are mistaken; otherwise, I would gladly help with a few more rewrites.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -23,7 +23,7 @@ The recommended way of setting your environment up is with [Anaconda](https://ww
 
 The project uses Python 3 (specified in the `python=3` bit of the commands below). [Download Anaconda](https://www.continuum.io/downloads) and create an _environment_ for the project.
 
-The Python version in Anaconda doesn't have to be 3 as `conda` environments manage other versions as well. But if you're going to do a fresh install we recommend Python 3 — it's newer and it makes no sense to have Python 2 as your default Anaconda Python.
+You can use either Anaconda 2 and Anaconda 3, since `conda` environments manage Python versions. But if you're going to do a fresh install, we recommend Anaconda 3 — it's newer and it makes no sense to have Python 2 as your default Anaconda Python.
 
 ```console
 $ cd serenata-de-amor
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -23,7 +23,7 @@ The recommended way of setting your environment up is with [Anaconda](https://ww
 
 The project uses Python 3 (specified in the `python=3` bit of the commands below). [Download Anaconda](https://www.continuum.io/downloads) and create an _environment_ for the project.
 
-The version of Anaconda doesn't have to be 3 as `conda` environments manage Python version as well. But if you're going to a fresh install we recommend Anaconda 3 — it's newer and it makes no sense to have Python 2 as your default Anaconda Python.
```
cuducos: Can you rollback this line, please? I mean… your version is way less precise what what we had:

* Both Anaconda and Anaconda3 can manage Python 2 and Python 3 environment under the hood, so I prefer _the version of Anaconda doesn't have to be 3_ instead of _the Python version in Anaconda doesn't have to be 3_;
* `coda` environments can use different Python versions, so I prefer _`conda` environments manage Python version as well_ instead of _`conda` environments manage other versions as well_);
* Actually we recommend Anaconda3 (and Python 3 is managed by the `conda` environment, so I prefer _we recommend Anaconda 3_ instead of _we recommend Python 3_.

CONTRIBUTING.md
```diff
@@ -102,7 +102,7 @@ Beyond that we have four big directories with different purposes:
 |-----------|---------|-------------|
 | **`develop/`** | This is where we _explore_ data, feel free to create your own notebook for your exploration. | `[ISO 8601 date]-[author-username]-[2-4 word description].ipynb` (e.g. `2016-05-13-anaschwendler-air-tickets.ipynb`) |
 |**`report/`** | This is where we write up the findings and results, here is where we put together different data, analysis and strategies to make a point, feel free to jump in. | Meaningful title for the report (e.g. `Transport-allowances.ipynb` |
-| **`src/`** | This is where our auxiliary scripts lies, code to scrap data, to convert stuff etc. | Small caps, no special character, `-` instead of spaces. |
```
cuducos: Actually _code to scrap data, to convert stuff_ is our _auxiliary scripts_. The enumeration after the comma was more expanding the meaning than adding stuff to _auxiliary_ script. If you think the writing is too unclear the way it is, would you mind replacing your _as well as_ by  _in other words_?

marceldalagnol: Sure! I understood they were separate things, but this makes more sense. Does the following work?
_This is where our auxiliary scripts lie: code to scrap data, to convert stuff etc._

CONTRIBUTING.md
cuducos: > You can use either Anaconda 2 and Anaconda 3, since conda environments manage Python versions. But if you're going to do a fresh install, we recommend Anaconda 3 — it's newer and it makes no sense to have Python 2 as your default Anaconda Python.

👍 Go for it!

CONTRIBUTING.md
cuducos: > This is where our auxiliary scripts lie: code to scrap data, to convert stuff etc.

+1 Hell yeah

cuducos:
Good that things got clearer — and probably it will get clearer for our readers once we merge your suggestions. Just commit the versions suggested above and LGTM 🎉

Issue #17, opened 2016-08-21
Find expenses made with companies of relatives


Irio:
According to "ATO DA MESA Nº 43, DE 21/5/2009 Art. 4º § 13", no expense can be reimbursed if made with companies where the applicant (politician) is owner, partner or relative (maximum of third degree).

g4brielvs:
I am interested to work on this issue. I am going to start by doing a descriptive analysis as simply as looking for common surnames and I see how far I can get.

jonasporto:
Hey guys, I did try analyse manually this case: 

http://m.congressoemfoco.uol.com.br/noticias/deputado-gasta-r-189-mil-da-cota-para-abastecer-no-posto-do-irmao/

So I did the following steps:

1. I did need to find the fullname of deputy "Wellington Roberto" and achieve this in: http://www.camara.leg.br/internet/deputado/dep_Detalhe.asp?id=5830395

2. So I would like to find out all of companies that him is involved as partner:
http://www.consultasocio.com/q/sa/jose-wellington-roberto 
There I found these names: Jose Edvan Roberto, Milena Roberto De Assis, Alto Branco Participacoes S/S Ltda, Walkiria Henriques Roberto(That I judge that all partner are part of family of deputy Jose Wellington Roberto).

3. So I did research for "Jose Edvan Roberto" and bazinga!!!! 
http://www.consultasocio.com/q/sa/jose-edvan-roberto
This guy(brother of wellington roberto) owns "Posto de Combustivel Shopping Ltda" company with all Roberto's family.

I think this is a try to obfuscate degree of family with partnership distribuited in nested companies.

Maybe this steps could be a way to find more cases like this.

Any thoughts?

pmargreff:
Hey, you can use [ConsultaSocio](http://www.consultasocio.com/) to search the relatives companies.

thiagobrito:
Hello,

I created a notebook and a pull request in order to close this issue.

The pull request link:
https://github.com/datasciencebr/serenata-de-amor/pull/168

The notebook link (in my own repository yet):
https://github.com/thiagobrito/serenata-de-amor/blob/nepotism-spends/develop/2016-12-15-thiagobrito-nepotism-spends.ipynb

bahbbc:
Hello! The main objective here is to generalize the manual analysis from @jonasporto? (Or try to find other ways?

cuducos:
> The main objective here is to generalize the manual analysis from @jonasporto?

[...]

PR #168, opened 2016-12-15
Find expenses made with companies of relatives

thiagobrito:
This pull request tries to close the issue #17.

I created this notebook in order to show all the spends that are relative to family companies spends.

In order to collect the database you must use the files:

- src/fetch_business_partners_info.py: This file collects all the information from ConsultaSocio.com using Google Cache and the website (until you get locked out).

Because takes a long time and work to download all ConsultaSocio.com and fetch all CNPJ information, I'll put my database as an attachment in this pull request and you should extract these files in ./data/ directory.

[data.zip](https://github.com/datasciencebr/serenata-de-amor/files/654696/data.zip)

This is my first contribution and hope to have a lot more and more. I see a lot of space to improve some areas of the code and I'll work on this soon.


src/fetch_business_partners_info.py
```diff
@@ -0,0 +1,104 @@
...
+        print('(%d/%d) %s' % (id, total_count, name), sep=' ', end='\n')
+        if name in result:
+            collected_count += 1
+        else:
+            default_info = {'partners': [], 'business': []}
+            if not original_site_blocked:
+                info = extract_info_from_page(make_consulta_socio_url(name), default_info, driver,
+                                              'Uso — ConsultaSocio.com')
+                original_site_blocked = info is None
+            if info is None:
```
cuducos: You could change these two lines (87-88) to compare `info` just once:

```python
if not info:
    original_site_blocked = True
    …
else:  # later on maybe `else` instead `if info` would work
    …
```

src/fetch_business_partners_info.py
```diff
@@ -0,0 +1,104 @@
...
+                info = extract_info_from_page(make_google_cache_url(name), default_info, driver, 'error')
+            if info:
+                result[name] = info
+                open(output_file_path, 'w').write(json.dumps(result))
```
cuducos: Is it possible to use what we have been using so far? I mean basically two things:

* Write the final output in CSV (tabular data) instead of JSON ([example](https://github.com/datasciencebr/serenata-de-amor/blob/master/src/fetch_cnpj_info.py))
* Compress the final database with `.xz` (if you're using tabular data and Pandas it's a matter of passing an extra argument, as in the link from the example above; if you stick with JSON, just use `lzma.open` instead of `open`)

src/fetch_business_partners_info.py
```diff
@@ -0,0 +1,104 @@
...
+    # Collect the information if there is no error
+    if error_msg.lower() not in soup.title.text.lower():
+        for partner in soup.findAll('span', attrs={'class': 'socio'}):
+            info['partners'].append(partner.text)
+        for company_info in soup.findAll('p', attrs={'class': 'number'}):
+            info['business'].append(company_info.find('span', {'class': 'text'}).text)
+        try:
+            info['cache_info'] = soup.find('div', attrs={'id': 'google-cache-hdr'}).text
+        except:
+            info['cache_info'] = str(datetime.datetime.now())
```
Irio: I see that `cache_info` may contain a string (as `'This is Google's cache of http://www.consultasocio.com/q/sa/nilson-aparecido-leitao. It is a snapshot of the page as it appeared on Dec 10, 2016 16:47:53 GMT. The current page could have changed in the meantime. Learn moreFull versionText-only versionView sourceTip: To quickly find your search term on this page, press Ctrl+F or ⌘-F (Mac) and use the find bar.'` or a datetime. If the idea of storing the Google Cache info is for having an option to go back and look for bugs in the future (good practice btw), can you create a new column `collected_at` or something like this, present in all the records, collected or not from the cache?

thiagobrito: Of course man! I'll work on this!

Irio:
I opened the `2016-12-13-congressperson-business-partners.json` file you attached and found a something weird. The `cache_info` attribute contains the URL for a non-related person in consultasocio.com. Is this a problem in the data collection itself or just in the outputted json?

Here's how to reproduce it. http://www.consultasocio.com/q/sa/nilson-aparecido-leitao appears for "PAULO CESAR DA GUIA ALMEIDA", for instance.
https://gist.github.com/56b4b5a72fb1f242b45896506ca3e75a

src/fetch_business_partners_info.py
```diff
@@ -0,0 +1,104 @@
...
+    driver.get(url)
+    driver.switch_to.default_content()
+    soup = BeautifulSoup(driver.page_source, 'html.parser')
+
+    # Somethimes google can ask for captcha, so you should type it!
+    if 'To continue, please type the characters below:' in soup.text:
+        while 'To continue, please type the characters below:' in soup.text:
+            driver.switch_to.default_content()
+            soup = BeautifulSoup(driver.page_source, 'html.parser')
+            time.sleep(1)
```
cuducos: Sorry about this last comment, I've talked to Irio but forgot to edit it before sending the review. Anyway: even if it's not the ideal it's not a big issue to depende on human interaction by now. Just a few questions: in your tests, how many times does it asks for the CAPTCHA (no problem if it's just a guess)? How long does it take to run it (the full dataset)? Cheers

carlosbaia:
Very good work! Congrats @thiagobrito 
Unbelievable, only with one source of data, you catched all this suspect transactions, imagine with we put all the data together. 
Man, I start thinking all the problems in Brazil are fault from this criminals.
I believe is valid send a notification for all of them, including for who are doing transaticion with partners, we know that's not illegal yet, but with this, they will know somebody is watching them.

src/fetch_business_partners_info.py
thiagobrito: In this case, with google captcha, it's hard to automate. It's the only way that I found in order to fix it.

src/fetch_business_partners_info.py
```diff
@@ -0,0 +1,104 @@
...
+    if error_msg.lower() not in soup.title.text.lower():
+        for partner in soup.findAll('span', attrs={'class': 'socio'}):
+            info['partners'].append(partner.text)
+        for company_info in soup.findAll('p', attrs={'class': 'number'}):
+            info['business'].append(company_info.find('span', {'class': 'text'}).text)
+        try:
+            info['cache_info'] = soup.find('div', attrs={'id': 'google-cache-hdr'}).text
+        except:
+            info['cache_info'] = str(datetime.datetime.now())
+        print(info)
```
thiagobrito: Yeah, it's for debugging purposes, but sometimes it's good to know what's going on because this process is really slow.

src/fetch_business_partners_info.py
```diff
@@ -0,0 +1,104 @@
...
+        return info
+
+
+def fetch_business_infos():
+    file_base_name = '{}-congressperson-business-partners.json'.format(datetime.date.today().strftime('%Y-%m-%d'))
+    output_file_path = os.path.join(os.path.abspath('../data'), file_base_name)
```
thiagobrito: I think it's better to get the .py file path and use it to calculate the path of the data directory. 

The way that you showed in the example find out the data directory as a subdirectory of src file. I don't think it's a good idea.

src/fetch_business_partners_info.py
```diff
@@ -0,0 +1,104 @@
...
+
+    # Load cached data if file exists
+    result = {}
+    if os.path.exists(output_file_path):
+        result = json.loads(open(output_file_path, 'r').read())
+
+    # Collect the information from all congress person
+    civil_names_df = pd.read_csv('../data/2016-12-07-congressperson-civil-names.xz')
```
thiagobrito: Ok, ok... good one ! :-)

thiagobrito:
@Irio @cuducos there is a ton of work here ! 

Thank you guys for your reviews... I'll fix it and try again! :)

src/fetch_business_partners_info.py
```diff
@@ -0,0 +1,104 @@
...
+import datetime
+import time
+import pandas as pd
+from unicodedata import normalize
+from selenium import webdriver
+from bs4 import BeautifulSoup
+
+
+def selenium_driver_path():
+    for path in ['chromedriver.exe', 'chromedriver']:
```
thiagobrito: > This adds decencies to the project that should be considered in the dev environment set up the process. I'm not a Selenium expert, but maybe we might expand this list so it will work with any of the major browsers, not only Chrome. Does that make sense?

Yeah, this makes sense and also I'll figure out how to download the driver in order to setup the environment automatically. But I'll make that the user must change the browser if want something different than Google Chrome and no Internet Explorer, please. :-)

> Maybe you can also include a warning for Docker users that this will not work on Docker (Docker users will have to download the database, which is not a bid deal TBH).

How can I do this?

> BTW do we really need Selenium? Does requests.get work in this case? I guess it wouldn't work because of the CAPTCHA, is that right?

When trying to collect information from Google Cache, that's the only way that I found to extract the information and insert the captcha when needed. But when collecting directly from ConsultaSocio.com it's better with requests. I'll fix that.

cuducos: I'd wouldn't say that users _need_ to be able to choose, but we must make sure that if someone doesn't have Chrome (or maybe Firefox) the script will try to find an alternative (even if the alternatives are restricted to Chrome and Firefox if it's difficult to address Safari, IE, whatever). 

About the Docker error message: I'd add a line to the documentation (of the script itself within `CONTRIBUTING.md`) and also in the `raise` threw by this method here.

Finally, CAPTCHA: I kind of imagined that scenario. But… how many times (from your  experience) one has to enter the CAPTCHA to go through the whole script? And how long dos the full process take?

src/fetch_business_partners_info.py
cuducos: Don't worry about that. Just give us (maybe add to the documentation) a note to warn possible users about how cumbersome it is.

src/fetch_business_partners_info.py
cuducos: I got your point but in Python it is a common practice to run scripts form the root of the project. Beyond that all other files from `src/` behave like that. It's a kind of standard for this (and most of Python) projects.

If you can implement a way that builds the path from the real location of the `.py` it's fine, it would work with `python fetch_business_partners_info.py` and `python data/fetch_business_partners_info.py`— otherwise can you make your script coherent with all its future buddies from `src/`?

src/fetch_business_partners_info.py
```diff
@@ -0,0 +1,104 @@
...
+        name = normalize_name(name).upper()
+
+        print('(%d/%d) %s' % (id, total_count, name), sep=' ', end='\n')
+        if name in result:
+            collected_count += 1
+        else:
+            default_info = {'partners': [], 'business': []}
+            if not original_site_blocked:
+                info = extract_info_from_page(make_consulta_socio_url(name), default_info, driver,
+                                              'Uso — ConsultaSocio.com')
```
thiagobrito: In this case, it's because when ConsultaSocio blocks our IP it redirects to a page that has this title "Uso - ConsultaSocio.com". But I'll change the name of the variable in order to fit with the purpose better.

src/fetch_business_partners_info.py
thiagobrito: @cuducos Is possible to create CSV using 1 to N items? 

For example, in this database, one congressman has one or more partners and companies. I really don't know, and didn't find any elegant solution in order to put this data that is better than using JSON format.

Any idea?

thiagobrito:
Hi, I made almost all code modifications asked, only the creation of CSV that I'm trying to solve yet as I told to @cuducos above.

I'll work to improve the data analysis now in order to fix @Irio issues.

src/fetch_business_partners_info.py
cuducos: > Is possible to create CSV using 1 to N items?

There are two strategies. One is exemplified in `YYYY-MM-DD-companies.xz` and the other one in `YYYY-MM-DD-congressperson_relatives.xz`. 

If you have a lot of different columns (eg `YYYY-MM-DD-companies.xz`) you can create a very large amount of columns, even if most of them are blank. In the companies dataset each company has around 20 columns with regular data (name, tarde name, address, etc.), plus 200 other columns for `secondary_activity` and `secondary_activity_code` (ie `secondary_activity_1` and `secondary_activity_1_code`, `secondary_activity_2` and `secondary_activity_2_code`, `secondary_activity_3` and `secondary_activity_3_code` up to 100).

If you don't have too much basic info, let's say, congressperson name, congressperson document, and just want to add a list of relative for each one of them, you just repeat lines (eg `YYYY-MM-DD-congressperson_relatives.xz`):

```csv
congressperson_document, congressperson_name, relative
42, Fulano de Tal, Ciclano
42, Fulano de Tal, Beltrano
42, Fulano de Tal, Etc
```

I think this is the best option for you.

Yet you can fit data in two columns: one dataset with the descriptive data (eg. congressperson name and document, an unique ID), and a second dataset with this unique ID (repeat lines as much as needed) and the extra info — both datasets are linked in a foreign key strategy. Works, but adds extra steps to analyze it later.

jtemporal:
hey @thiagobrito I really want to merge this one! Is there anything I can do to help you on what is missing?

develop/2016-12-15-thiagobrito-nepotism-spends.py
```diff
@@ -0,0 +1,358 @@
...
+def save_cnpj_cache():
+    global cache
+    open(cache_path, 'w').write(json.dumps(cache))
+    
+def extract_cnpj_board(cnpj):
+    global cache
+    result = []
+    try:
+        if cnpj not in cache:
+            url = 'http://receitaws.com.br/v1/cnpj/%s' % cnpj
```
jtemporal: You can also use a more recent one that doesn't have geolocation: `2017-05-21-companies-no-geolocation.xz`

cuducos:
Closed by inactivity

Issue #143, opened 2016-11-21
Use serenata_toolbox whenever possible


Irio:
https://github.com/datasciencebr/serenata-toolbox was created to keep code shared between different Serenata de Amor projects. If the script is already there, the code should be removed from here and used from the package.

vmesel:
I think we should add this to the README file so we can ask people to use it, letting it on an issue is not so accessible as it must be.

cuducos:
Almost agree, but CONTRIBUTING.md is where this info is gonna be added IMHO

fgrehm:
I was willing to give this a shot but it seems that the package [is not on pypi yet](https://pypi.python.org/pypi/serenata_toolbox/), any recommendations on how to leverage that here? (git submodules, `wget`, something else...)

cuducos:
We can install directly from GitHub [with the -e argument](
https://github.com/datasciencebr/rosie/blob/master/requirements.txt) ; )

anaschwendler:
Talking with @Irio what's missing:
- Add the serenata-toolbox to requirements.txt
- We need to use serenata-toolbox in the notebooks

@jtemporal

jtemporal:
feedback:
- [x] Add the serenata-toolbox to requirements.txt
- [x] Make sure that serenata toolbox works on the notebooks
- [ ] Remove from src all code that is already in the toolbox
- [ ] Check/Update documentation

PR #172, opened 2016-12-27
Analysis of meals on speech days

fgrehm:
_Related to GH-137_

This also has the the bonus of setting up [serenata_toolbox](https://github.com/datasciencebr/serenata-de-amor/issues/143). It is only used for this notebook ATM but I can look into updating other stuff to make use of it at another day this or next week 🤘


requirements.txt
```diff
@@ -1,2 +1,4 @@
+-e git+https://github.com/datasciencebr/serenata-toolbox.git@6126436#egg=serenata-toolbox
```
Irio: @cuducos You [recently removed](https://github.com/datasciencebr/rosie/commit/9414eb3195ea7c4b9c8d9b8d073f3bbd078482e2) the `-e` from Rosie. Was for not saving it inside the project folder, right?

If that's the case, can you change for the version after @cuducos' commit?

cuducos: Yep. `-e` means editable mode, so it copy the package source code to your project. We don't need that here ; )

fgrehm: Sure! I'll get that changed later on today 👍

requirements.txt
```diff
@@ -1,2 +1,4 @@
+git+https://github.com/datasciencebr/serenata-toolbox.git@6126436#egg=serenata-toolbox
```
fgrehm: @Irio @cuducos I've removed the `-e` but was wondering if you guys had any thoughts on keeping it locked to a version like I did here? I like that it makes things reproductible and makes it clear when we expect people to update the package on their local installations but I noticed you guys didn't do it on rosie 💭

cuducos: Very very personal opinion here: locking the version surely is a good thing, but I think it should be done in a neat way. It's a good point you raise here, I don't think we've pondered on it in this context… Anyway: a neat way would be to use GitHug tags, or (I prefer) uploading it to PyPI.

fgrehm: yeah, uploading to pypi would be ideal, but a git tag might be the easiest way to move forward for now 💭

cuducos: @irio: Any ideas? If the silence remains (_drama_ 👑 ) I' upload it to PyPI next week hahaha…

Irio: @cuducos  :shipit:

fgrehm:
@Irio @cuducos since master already have serenata-toolbox, I guess this can be merged now

PR #179, opened 2017-01-16
Docker dev environment tweaks

fgrehm:
Follow up of a discussion over Telegram, this PR:

- Locks the docker image to a known to work tag for reproducibility
- Updates contributing info with the size of the image in use
- Disables authentication altogether (the OP on Telegram mentioned he had issues with the token that was generated for him). I don't think this is an issue since this is not meant to be used in production AFAIK.

/cc @cuducos

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -62,7 +62,7 @@ $ source /usr/local/var/pyenv/versions/anaconda3-4.1.1/bin/activate serenata_de_
 
 You can user [Docker](https://docs.docker.com/engine/installation/) and [Docker Compose](https://docs.docker.com/compose/install/) to have a working environment:
 
-1. Start the environment (it might take a while, the base image has 5.3GB and we also pull in lots of dependencies): `$ docker-compose up -d`
+1. Start the environment (it might take a while, the base image has 5.8GB and we also pull in lots of dependencies): `$ docker-compose up -d`
 1. Create your `config.ini` file from the example: `$ cp config.ini.example config.ini`
 1. Run the script to download data and other useful files: `$ docker-compose run --rm jupyter python src/fetch_datasets.py`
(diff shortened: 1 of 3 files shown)


CONTRIBUTING.md
```diff
@@ -62,7 +62,7 @@ $ source /usr/local/var/pyenv/versions/anaconda3-4.1.1/bin/activate serenata_de_
 
 You can user [Docker](https://docs.docker.com/engine/installation/) and [Docker Compose](https://docs.docker.com/compose/install/) to have a working environment:
 
-1. Start the environment (it might take a while, the base image has 5.3GB and we also pull in lots of dependencies): `$ docker-compose up -d`
+1. Start the environment (it might take a while, the base image has 5.8GB and we also pull in lots of dependencies): `$ docker-compose up -d`
```
cuducos: 💅 We really should use _roughly 5Gb_ hahaha…

docker/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM jupyter/datascience-notebook:latest
+FROM jupyter/datascience-notebook:fa77fe99579b
```
cuducos: Is there a specific reason to use a fixed tag instead of a version, or simply the latest?

fgrehm: As described on the PR description, same reason as we have versioned datasets: reproducibility 😄 as it just happened, we don't want a potential future contributor to have trouble with his / her environment because some change that was introduced in a newer release.

CONTRIBUTING.md
fgrehm: I think that's good for now as it is _roughly_ accurate (real size is 5.82 :P), if we need to change it again I agree we should add that :)

docker/Dockerfile
gomex: I like it 👍

Issue #167, opened 2016-12-14
Allow /src scripts to receive data files as command line arguments


Irio:
The majority of scripts located in process datasets in hard coded locations, like `data/cnpj-info.xz` in [src/fetch_cnpj_info.py#L10](https://github.com/datasciencebr/serenata-de-amor/blob/a599fe339b4cd13a2b9c56e4f2521caa16f55973/src/fetch_cnpj_info.py#L10). Given a lot has changed since their creation, we expect them to receive data paths as command line arguments, as in `python src/fetch_cnpj_info.py data/2016-12-06-cnpj-info.xz`.

marcusrehm:
Hi @Irio ! I'm working exactly that. I spoke with @cuducos about issue #67 that I also need to use **fetch_cnpj_info** with the amendments' dataset .

Actually, I was thinking about passing the column with the CNPJ's as a parameter also, so the call would be *src/fetch_cnpj_info.py './data/2016-12-06-cnpj-info.xz' 'cnpj'*. What do you think?

marcusrehm:
Hi @Irio 

I did the refactoring in [fetch_cnpj_info.py](https://github.com/marcusrehm/serenata-de-amor/blob/issue-67/src/fetch_cnpj_info.py#L125:L128) now we can call it passing filename and column with CNPJ's.

cuducos:
Sorry if I missed anything… but I don't think passing the column we need is what I had in mind. When I read OP's description I think he meant the file to be load (for example `2016-08-08-companies.xv` or `2016-12-06-companies.xz`). Also I'm afraid that's not what I meant in #67, @marcusrehm (but I'm gonna clarify that on the issue page).

marcusrehm:
@Irio @cuducos I think this one was resolved with PR #185 as in https://github.com/datasciencebr/serenata-de-amor/blob/master/src/fetch_cnpj_info.py#L159 .

cuducos:
@marcusrehm PR #185 addresses this issue when it come to company scripts, but not for all scripts inside `src/` directory ; )

marcusrehm:
yes, @cuducos , you're right! =)

martini97:
[...]

Issue #67, opened 2016-09-15
Looking for corruption on the Federal Budget 


franklinbaldo:
The Brazilian Constitution allows each parliamentary allocate a portion of the federal budget for a specific purpose. But there is a problem because the law also allowed the parliamentary indicate the institution (NGO, Association, Foundation, public agency) that will receive the money. This creates a major risk of embezzlement, if the money is intended to entities controlled by the Parliament itself.

The federal government publishes the list of entities that received funds in this way. This list indicates which entity received the money, what she should do and what was the congressman who was the author of the amendment. address http://portal.convenios.gov.br/images/docs/CGSIS/csv/siconv_emenda.csv.zip

We could build a tool to check the reputation of such entities. This information would indicate higher risk of corruption.

We can get the information about the reputation from various sources: protests because of debts for this CNPJ, jundiciais actions against authority (sites of the courts, JusBrasil), criminal actions against the leaders (courts sites), leaders of donations to campaign parliamentary (TSE) and others.
# Português

A constituição brasileira permite que cada parlamentar destine uma parte do orçamento federal para um objetivo específico. Mas existe um problema porque a Lei também permite que o parlamentar indique a instituição (ONG, Associação, Fundação, órgão público) que irá receber esse dinheiro. Isso gera um grande risco de desvio de dinheiro, se o dinheiro for destinado a entidades controladas pelo próprio parlamentar. 

O governo federal divulga a lista de entidades que receberam verbas dessa forma. Essa lista indica para qual entidade recebeu o dinheiro, o que ela deveria fazer e quem foi o parlamentar que foi autor da emenda.  Endereço: http://portal.convenios.gov.br/images/docs/CGSIS/csv/siconv_emenda.csv.zip

Nós poderíamos construir uma ferramenta que verifique a reputação dessas entidades. Essa informação indicaria emendas com alto risco de corrupção.

Podemos obter as informações sobre a reputação a partir de várias fontes: protestos em razão de dívidas cíveis (buscar pelo CNPJ em sites como http://www.ieptb.com.br/), ações judiciais contra entidade (sites dos tribunais, jusbrasil), ações civis e criminais contra os dirigentes (sites de tribunais), doações de dirigentes a campanha do parlamentar ou do partido (TSE) e outras.

cuducos:
Awesome, @franklinbaldo!

IMHO the steps to bring this in to our project would be something like these:

[...]

PR #185, opened 2017-01-27
Issue #67 - Looking for corruption on the Federal Budget

marcusrehm:
Related to issue #67 and maybe #167 .


.gitignore
```diff
@@ -4,3 +4,5 @@
 __pycache__/
 config.ini
 data
+.idea/
+serenata-de-amor.iml
```
cuducos: Just out of curiosity — what is this file about?

marcusrehm: Sorry @cuducos ! At beginning I was using IntelliJ IDE and this file is generated automatically with some IDE project's properties.

Do you want me to remove that?

cuducos: No way. more people can use IntelluJ IDE too ; )
It was just my curiosity…

marcusrehm: Ok! :)

PR #186, opened 2017-02-05
Add dockerignore to prevent upload of data dir and other files to docker engine on builds

fgrehm:
This prevents files under `data/` and other stuff that is gitignored from being added as the context of docker builds.

This is a problem since (for example) I've just downloaded 5Gb worth of PDF receipts into that dir and when I had to rebuild my env the `docker-compose` command became unresponsive and builds where taking waaaaay too long to complete (I really wish docker defaulted to looking up gitignores but that's not the case)

/cc @cuducos @gomex

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,8 @@
+*.pyc
+.ipynb_checkpoints
+.python-version
+__pycache__/
+config.ini
+data
+.idea/
+serenata-de-amor.iml


.dockerignore
```diff
@@ -0,0 +1,8 @@
+*.pyc
+.ipynb_checkpoints
+.python-version
+__pycache__/
+config.ini
```
cuducos: Shouldn't `config.ini` be available inside the container?

fgrehm: Doesn't really matter, the current dir will be mounted inside the container and people are supposed to `cp config.ini.example config.ini` anyways ;)

cuducos:
@irio is working in a new Docker architecture. Just tagging him here so he can take that into account, ok?

PR #187, opened 2017-02-07
[WIP] Family Allowance data collection

arnaldocorreiafilho:
In relation to question # 77 the data of the family bag is already working the code that downloads all the files will do a pull requeste and see if everything is ok.


anaschwendler:
Hi @arnaldocorreiafilho!    
I didn't get what do you want to do with this PR, sorry. Is that a work in progress and you working on it, or there should be more code into it?

Thank you!

conda_requirements.txt
```diff
@@ -7,3 +7,5 @@ pandas>=0.18.1
 scikit-learn==0.18.1
 seaborn==0.7.1
 ipython==5.1.0
+python-wget=3.2
```
cuducos: This package doesn't look like to be available in PyPI:

```console
$ pip install python-wget
Collecting python-wget
  Could not find a version that satisfies the requirement python-wget (from versions: )
No matching distribution found for python-wget
```

cuducos: Also there's a typo: it should have a double equal (`==`) not a single one (`=`).

arnaldocorreiafilho: This lib is a wget-to-python catch that can be found on this site https://anaconda.org/bioconda/python-wget. Is used to do file downloads.

cuducos: Ooops… my bad. I shouldn't have tried it with `pip` ; )

Just out of curiosity, why not using native [`urlretrieve`](https://docs.python.org/3.6/library/urllib.request.html#urllib.request.urlretrieve) that already has a straightforward API?

arnaldocorreiafilho:
The data is already being downloaded, which should have more in this module since I am a beginner in the project and contribute with free software projects. At first I thought that the principle would only be to make the script that downloaded the data of the beneficiaries of bolsa famiia.

cuducos:
Maybe you've forgotten to commit the script itself — check the [*Files changed* tab](https://github.com/datasciencebr/serenata-de-amor/pull/187/files): your PR just adds a couple of lines to `conda_requirements.txt` ; )

Besides that, a couple of hints to enhance your contribution:

* If you're adding a new script, don't forget to document it in `CONTRIBUTING.md`;
* If you're adding a new dataset that takes time to download/generate/scrap/etc, share a link to the dataset you already have, so we can check data and make it available in our S3 server ; )

arnaldocorreiafilho:
I finished the script that downloads the data from the family bag. Do I need to create a new pr?

arnaldocorreiafilho:
In regards to putting the data in a file. I am having difficulties since the files are very large when uncompressed they reach 1G and are 12 for each year so it will take a lot to gather all in one file. Unless when you finish the downloads Of the files compress all in one zip but keep them separate inside the compressed zip. Any suggestion ?

cuducos:
Maybe we can iterate over the monthly files without having them loaded in memory, and write lines to a file that gathers all data also without having it loaded in memory. Pandas or [Dask](https://github.com/datasciencebr/serenata-de-amor/issues/15) can help the reading, and the attach mode (`a`) can be helpful when writing the file… does that make sense?

arnaldocorreiafilho:
It makes sense, yes. But to gather all the files will take a lot of time. I spent the whole weekend running this code and was missing many files to be added. The memory problem has been solved, however, because large files take too long to be concatenated and compromise the installation time of the environment.
The current code is the one.


import datetime
import warnings
import requests
import zipfile
from urllib.request import urlretrieve
import os
import pandas
import fileinput
from sqlalchemy import create_engine
import sys
import shutil

STARTING_YEAR = 2011

NOW = datetime.datetime.now()

SERVER = 'http://arquivos.portaldatransparencia.gov.br/'
PATH = 'downloads.asp?a={}&m={:02d}&consulta=BolsaFamiliaFolhaPagamento'

BASE_DATA_DIR = '../data/'
TEMP_DATA = "../data/temp_data/"


COLLUNM_DATA = ['UF', 'CODIGO', 'SIAFI_MUNICIPIO', 'NOME_MUNICIPIO', 'CODIGO_FUNCAO', 'CODIGO_SUBFUNCAO', 'CODIGO_PROGRAMA', 'CODIGO_ACAO',
                 'NIS_FAVORECIDO', 'NOME_FAVORECIDO', 'FONTE_FINALIDADE', 'VALOR_PARCELA', 'MES_COMPETENCIA']


def dlProgress(count, blockSize, totalSize):
  percent = int(count*blockSize*100/totalSize)
  sys.stdout.write("\r" + "progress" + "...%d%%" % percent)
  sys.stdout.flush()



def urls():
    for year in range(STARTING_YEAR, NOW.year + 1):
        for month in range(1, 13):
            if datetime.datetime(year, month, NOW.day) <= NOW:
                url = SERVER + PATH.format(year, month)
                req = requests.head(url)

                # looks like some devs are not familiar with 404 status
                if req.headers.get('Content-Type') == 'application/x-download':
                    yield (url,year,month)

                else:
                    msg = 'Data from {:02d}/{} could not be located at: {}'
                    warnings.warn(msg.format(month, year, url))

def concat_csv_file(path_file):
    chunksize = 2000

    df = pandas.DataFrame(columns=COLLUNM_DATA)
    i=0
    j=0
    for df in pandas.read_csv(path_file, chunksize=chunksize, iterator=True,sep='\t',encoding='latin-1'):
         df = df.rename(columns={c: c.replace(' ', '') for c in df.columns})
         df.index += j
         i += 1
         if not os.path.isfile(BASE_DATA_DIR+'filename.csv'):
             df.to_csv(BASE_DATA_DIR+'filename.csv')
         else:  # else it exists so append without writing the header
             df.to_csv(BASE_DATA_DIR+'filename.csv', mode='a', header=False)
         j = df.index[-1] + 1



def download_datasets():
    os.makedirs(TEMP_DATA, exist_ok=True)
    for data in urls():
        url, year, month =  data
        print (url,year,month)
        filename = BASE_DATA_DIR +"file" +str(year) + str(month)+".zip"
        urlretrieve(url,filename,reporthook=dlProgress)
        zip_ref = zipfile.ZipFile(filename, 'r')
        print(zip_ref.namelist()[0])
        zip_ref.extractall(TEMP_DATA)
        concat_csv_file(TEMP_DATA+zip_ref.namelist()[0])
        zip_ref.close()



if __name__ == '__main__':
    download_datasets()
    shutil.rmtree(TEMP_DATA)

cuducos:
@arnaldocorreiafilho Many thanks for all the effort!

Would you mind committing the code to the PR — it's easier to give feedback in inline comments ; )

src/fetch_family_allowance.py
```diff
@@ -0,0 +1,91 @@
+import datetime
+import warnings
+import requests
+import zipfile
+from urllib.request import urlretrieve
+import os
+import pandas
+import fileinput
```
cuducos: Are we using this module? Can we delete it?

src/fetch_family_allowance.py
```diff
@@ -0,0 +1,91 @@
...
+from sqlalchemy import create_engine
```
cuducos: Are we using this module? Can we delete it?

src/fetch_family_allowance.py
```diff
@@ -0,0 +1,91 @@
+import datetime
```
cuducos: Would you mind organizing a bit this import section? In PEP 8 we have something like:

> Imports should be grouped in the following order:
> * standard library imports
> * related third party imports
> * local application/library specific imports
> 
> You should put a blank line between each group of imports.

We don't have to be strict, but this could be inspirational source ; )

src/fetch_family_allowance.py
```diff
@@ -0,0 +1,91 @@
...
+import sys
+import shutil
+
+STARTING_YEAR = 2011
+
+NOW = datetime.datetime.now()
+
+SERVER = 'http://arquivos.portaldatransparencia.gov.br/'
+PATH = 'downloads.asp?a={}&m={:02d}&consulta=BolsaFamiliaFolhaPagamento'
+outputfile = 'data_family_allowance.csv'
```
cuducos: All constants are in caps, can we have that one in caps too?

Also we don't seem to use `outputfile` alone: it's always `BASE_DATA_DIR + outputfile` — what about concatenating `BASE_DATA_DIR` to the output file constant?

src/fetch_family_allowance.py
```diff
@@ -0,0 +1,91 @@
...
+BASE_DATA_DIR = '../data/'
```
cuducos: This is not compatible with Windows (they use `\` instead of `/`). You can use the `os` native module to properly build file and directory paths in a compatible way. Just check how others scripts in the `src/` folder do it as a reference if you need it ; )

src/fetch_family_allowance.py
```diff
@@ -0,0 +1,91 @@
...
+TEMP_DATA = "../data/temp_data/"
```
cuducos: Can we use [`tempfile.TemporaryDirectory`](https://docs.python.org/3/library/tempfile.html#tempfile.TemporaryDirectory)?

src/fetch_family_allowance.py
```diff
@@ -0,0 +1,91 @@
...
+              'NomeFavorecido': 'FAVORED_NAME',
+              'Fonte-Finalidade': 'SOURCE_FINALITY',
+              'ValorParcela': 'PARCEL_VALUE',
+              'MêsCompetência': 'MONTH_COMPETENCE',
+          }
+
+def dlProgress(count, blockSize, totalSize):
+  percent = int(count*blockSize*100/totalSize)
+  sys.stdout.write("\r" + "progress" + "...%d%%" % percent)
+  sys.stdout.flush()
```
cuducos: We can replace these two lines with `print(…, end='/r')` — it's simpler and cleaner.

src/fetch_family_allowance.py
```diff
@@ -0,0 +1,91 @@
...
+             df.to_csv(BASE_DATA_DIR+outputfile)
+         else:  # else it exists so append without writing the header
+             df.to_csv(BASE_DATA_DIR+outputfile, mode='a', header=False)
+         j = df.index[-1] + 1
+
+
+def download_datasets():
+    os.makedirs(TEMP_DATA, exist_ok=True)
+    for data in urls():
+        url, year, month =  data
```
cuducos: These two lines could be just `for url, year, month in urls` ; )

src/url_family_bag.py
```diff
@@ -0,0 +1,29 @@
+import datetime
```
cuducos: Do we still need this file? It's logic is already in the other one.

src/fetch_family_allowance.py
```diff
@@ -0,0 +1,91 @@
...
+
+                # looks like some devs are not familiar with 404 status
+                if req.headers.get('Content-Type') == 'application/x-download':
+                    yield (url,year,month)
+
+                else:
+                    msg = 'Data from {:02d}/{} could not be located at: {}'
+                    warnings.warn(msg.format(month, year, url))
+
+def concat_csv_file(path_file):
```
cuducos: Why do we need this function? `urlretrive` already saves the file into the disk, right? The code is not clear, the lack of comments/docstrings and generci variables names (`i` and `j`) makes it really difficult to grasp it's idea.

Maybe a better strategy is to write three function and keep they simple:
1. `save_file` just wrapping `urlretrieve`
2. `unzip_file` (maybe merge it with the previous function, but anyway…) to uncompress the download file
3. `translate_headers` that would just read the first line and rewrite it

cuducos:
Closed by inactivity

Issue #189, opened 2017-02-15
New data about companies


allantorres:
I was searching about CNPJ APIs and i found a great beta API from Compras.dados.gov.br... 

http://compras.dados.gov.br/fornecedores/doc/fornecedor_pj/01645738000179

http://compras.dados.gov.br/fornecedores/doc/fornecedor_pj/01645738000179.json

This API show all information about the company and the contracts they have with the governament.

Irio:
Great finding @allantorres.

anapaulagomes:
I made a PR (#193) related to supplier's data collection. I wish contribute with some analysis related with the purchases and materials.

caduvieira:
We could use https://www.receitaws.com.br/ also. There is available a free API there.

cuducos:
> We could use https://www.receitaws.com.br/ also. There is available a free API there.

Why?

ReceitaWS has information about the company — [which we're already collecting](https://github.com/datasciencebr/serenata-de-amor/blob/19f7b603e0b396b1be27f62c901849c481d83cd4/src/fetch_cnpj_info.py#L71), but nothing on the contracts they have with the governament — so @allantorres's suggestion still has a strong point over ReceitaWS.

However, as mentioned #16 recently _now_ Receita WS is *also* publishing the partners of each company. It's a very very interesting to include in [`src/fetch_cnpj_info.py`](https://github.com/datasciencebr/serenata-de-amor/blob/19f7b603e0b396b1be27f62c901849c481d83cd4/src/fetch_cnpj_info.py#L71) ; )

caduvieira:
Oh ok. I got it now. There is another API in siconv for data of public money transfers. Some company could be of some congressperson parent. http://dados.gov.br/dataset/siconv

cuducos:
Closed by #193

PR #193, opened 2017-02-18
Fetch purchase's suppliers from compras.dados.gov.br

anapaulagomes:
Related with issue #189. The script collects the data of all suppliers related with purchases made by federal government. The purchases could be by a company or by a person. The API is brand new and ew have few info in this website: http://compras.dados.gov.br:8080/docs/lista-metodos-fornecedores.html.


src/fetch_purchases_suppliers.py
```diff
@@ -0,0 +1,40 @@
+import requests
+import pandas as pd
+
+
+BASE_DATA_DIR = './data/'
```
cuducos: I'd recommend using `os.path` to build the path so Python handles the `os.sep` and other compatibility issues (`/` for Linux/macOs and `\` for Windows, for example).

src/fetch_purchases_suppliers.py
```diff
@@ -0,0 +1,40 @@
...
+
+def fetch_suppliers():
```
cuducos: This is script doesn't seem like to memory consuming, but just in case, I would write `fetch_suppliers` as a generator:

```python
def fetch_suppliers():
    # no `raw_suppliers = []`
    …
    response = requests.get(…).json()  # actually this was your `first_response`
    yield response

    …
    for page in range(2, last_page):
        …
        yield response
```

What do you think?

src/fetch_purchases_suppliers.py
```diff
@@ -0,0 +1,40 @@
...
+    raw_suppliers = []
+    first_response = requests.get('http://compras.dados.gov.br/fornecedores/v1/fornecedores.json').json()
+    raw_suppliers = raw_suppliers + first_response['_embedded']['fornecedores']
+
+    last_page = first_response['count']
+    for page in range(2, last_page):
+        print('{}/{}'.format(page, last_page))
+        response = requests.get('http://compras.dados.gov.br/fornecedores/v1/fornecedores.json').json()
+        raw_suppliers = raw_suppliers + first_response['_embedded']['fornecedores']
```
cuducos: You're always appending `first_response` (not the `response` from the `for` loop)… is that right?

src/fetch_purchases_suppliers.py
```diff
@@ -0,0 +1,40 @@
...
+    return raw_suppliers
+
+
+def prepare_data(raw_data):
+    suppliers_attributes_columns = {'id': 'id', 'cnpj': 'cnpj', 'nome': 'name', 'ativo': 'active',
```
cuducos: This is quite difficult to read, what do you think about formatting it [like that](https://github.com/datasciencebr/serenata-de-amor/blob/master/src/clean_cnpj_info_dataset.py#L38-L59)?

src/fetch_purchases_suppliers.py
```diff
@@ -0,0 +1,40 @@
...
+    'id_porte_empresa': 'company_size_id', 'id_ramo_negocio': 'business_id', 'id_unidade_cadastradora': 'responsible_entity',
+    'id_cnae': 'cnae_id', 'habilitado_licitar': 'allowed_to_bid'}
+
+    df = pd.DataFrame(raw_data, columns = suppliers_attributes_columns.keys())
+
+    return df
+
+
+def save_csv(df):
+    df.to_csv(path_or_buf=BASE_DATA_DIR+'purchase_suppliers.xz', sep=',', compression='xz', encoding='utf-8', index=False)
```
cuducos: We have the practice of adding a date prefix to every dataset (`YYYY-MM-DD`) as a way of versioning our datasets (and consequentially our scripts and notebooks). Would you mind adding that?

src/fetch_purchases_suppliers.py
anapaulagomes: Much better! 😅

src/fetch_purchases_suppliers.py
anapaulagomes: Actually this doesn't make any sense... 😳  Fixing.

src/fetch_purchases_suppliers.py
anapaulagomes: Great idea! I don't have much experience with generators but I took a shot. Please give me some feedback!

src/fetch_purchases_suppliers.py
anapaulagomes: Fixed.

src/fetch_purchases_suppliers.py
anapaulagomes: Done.

cuducos:
Yay — many thanks 🎉

Issue #101, opened 2016-11-03
Plugin to use Neo4j within Jupyter notebooks


cuducos:
This is not an issue that will generate code for Serenata de Amor, but will create a while new package.

Using graph database will help in Serenata de Ampr us and I ran into @nicolewhite's repo that makes great dataviz using Neo4j inside Jupyter notebooks (check this [example](http://nicolewhite.github.io/neo4j-jupyter/hello-world.html) and the main [repo](https://github.com/nicolewhite/neo4j-jupyter)).

My idea (and via tweet Nicole's said she's fine with that) is to pack her code into a package instalable via `pip`. Adding the module to Serenata de Amor repo and we can use the dataviz in out notebook. That's the idea.

cuducos:
I started to adapt the code but I couldn't get back to it (it's **not** working). I have this work in progress though https://github.com/cuducos/neo4jupyter

Gusbru:
I managed how to do Neo4j works. I am preparing a documentation together with a jupyter notebook explaining which packages are necessary together with the "Hello world" program.

cuducos:
Many thanks, @Gusbru — that's very helpful ; )

vmesel:
@renanpalmeira can you help us?

RenanPalmeira:
@vmesel @cuducos I know a few plugins for neo4j with Jupyter, but never used. I have a simple project (a simple ORM) above the Neo4j with python https://github.com/RenanPalmeira/pyneo4j. Nothing prevents us from creating a plugin for jupyter.

About Neo4j in Serenata de amor, The perfect ideia because the government itself uses neo4j to look for irregularities (https://www.infoq.com/br/presentations/big-data-no-planalto-como-o-governo-minera-datasets).

I use Neo4j in production long ago, I can help in to model in Neo4j and in plugin jupyter.

cuducos:
If you can test any of these plugins, addind the instructions to
CONTRIBUTING.md and (maybe) an example Jupyter Notebook this issue is
closed ; )

marcusrehm:
Hey guys, I played with it for a while and created a branch [here](https://github.com/marcusrehm/serenata-de-amor/tree/neo4j).

I updated the [CONTRIBUTING.MD](https://github.com/marcusrehm/serenata-de-amor/blob/neo4j/CONTRIBUTING.md#using-neo4j-in-graph-analysis) to show how to put Neo4j running with Docker and created 2 notebooks using it.

[...]

Issue #174, opened 2017-01-04
Look for companies with congresspeople names


evilasiov:
Sometimes the trade company name is the owner's name, specially when it has just a single parter.  Searching for these companies (possibly filtering by `legal_entity` equals to `213-5 - EMPRESARIO (INDIVIDUAL)`) may bring good results.  Related with https://github.com/datasciencebr/serenata-de-amor/issues/17.

evilasiov:
Going to try to work on it.

cuducos:
Closed accidentally by unrelated commit from Rosie/Jarbas repos.

Issue #139, opened 2016-11-17
List of advisors of each congressperson


Irio:
Scrape from the Chamber of Deputies' site, [as suggested by Fernando Noronha](https://www.youtube.com/watch?v=_E6Iwm9MMKI&lc=z12yunpwvmjagnywb04cft2pxxjkzpv4fuk0k).

http://www2.camara.leg.br/transparencia/recursos-humanos/quadro-remuneratorio/consulta-secretarios-parlamentares/layouts_transpar_quadroremuner_consultaSecretariosParlamentares

It's a dependency of https://github.com/datasciencebr/serenata-de-amor/issues/138.

bruno-schmidt:
Hello guys! Did anybody tried to solve this issue yet?

I tried something last night. Please, take a look: https://github.com/bruno-schmidt/sda-advisors-scrapper

The code is a shame, I'm sorry. I'm beginning to learn all this stuff with javascript, so it might look ugly.  

Please let me know if it helps!

Important: it takes some (really some) time to scrap all the data! I didn't try it yet because I've been experiencing connection issues, but I will try again soon.

cuducos:
Many thanks, @bruno-schmidt.

If your issue is performance, bear in mind that you're making ~514 requests and yes, requesting, treating the response, writing data are steps that take time. That said my comments here are code-wise (not data wise, I haven't check that part yet).

From my tiny experience in this project multiprocessing helps, but what really enhances performance in scripts like yours  is parallelizing the requests (our first bottleneck isn't the processor, it is the amount of requests in our imaginary queue). I've been playing around with [grequests](https://github.com/kennethreitz/grequests) (that's not a typo, that's [gevent](http://www.gevent.org) + [requests](http://docs.python-requests.org/)). Maybe you can pack a group of requests and start them all in parallel. 

Multiprocessing workers are limited by the processor (usually 2 or 4, max. 8 for a common computer). But more than 8 requests can happen in parallel without being an hassle to the processor.

An alternative might be using [Scrapy framework](https://scrapy.org) — they optimize lots of things for performance (and their focus is surely scrapping data). But this might be too much: you're script is already in good shape, I wouldn't rewrite everything in the scrapy framework at this point.

Let me know if this makes sense. I might be available for pairing one of these days if you want — just drop a line.

bruno-schmidt:
Thank you @cuducos!

It really makes sense. I made some research again and rewrote some code using grequests as you suggested. It's running considerable faster!

[...]

PR #200, opened 2017-03-10
Plugin to use Neo4j within Jupyter notebooks

marcusrehm:
Plugin to use Neo4j within Jupyter notebooks #101


cuducos:
This is a really good contribution and I'm looking forward to see people using _graph_ in Serenata. Many thanks, @marcusrehm!

Before merging I'd like to suggest three things:

- [ ] I'd keep the one single Docker `yml`: let's enable Neo4j by default for everyone using Docker;
- [ ] This would make the `CONTRIBUTING.md` way simpler, just a few lines saying that Neo4j needs to be installed separately;
- [ ] In this sense of cleaning up `CONTRIBUTING.md` I would skip the details about `pip install` (it's a default procedure) and Neo4j instructions (maybe move it to the example notebook).

What do you think about it?

marcusrehm:
Hi @cuducos !

I agree with your comments, I think the CONTRIBUTING.MD got bigger, so I made some adjustments in Neo4j section as you suggested.

>I'd keep the one single Docker yml: let's enable Neo4j by default for everyone using Docker;

Done!

>This would make the CONTRIBUTING.md way simpler, just a few lines saying that Neo4j needs to be installed separately;

Most of steps to configure and test were moved to [2017-02-12-marcusrehm-neo4j-guide](https://github.com/marcusrehm/serenata-de-amor/blob/neo4j/develop/2017-02-12-marcusrehm-neo4j-guide.ipynb) notebook.

>In this sense of cleaning up CONTRIBUTING.md I would skip the details about pip install (it's a default procedure) and Neo4j instructions (maybe move it to the example notebook).

I putted the required packages inside default requirement.txt.

> I'm looking forward to see people using graph in Serenata

Yes! In fact, while developing this issue I was following others and I think using Neo4j in issues like #17, #18, #98, #139 and maybe #174 would be of a great help. =)

cuducos:
That's awesome! Thanks, @marcusrehm ; )

PR #203, opened 2017-03-18
Descriptive analysis of local transportation II

fabiocorreacordeiro:
This is Exploratory Descriptive Analisys of expendure with local
transportation. This notebooks is about subquota 'Fuels and lubricants'.
We found 24,480 suspicious reimbursement, a total of R$66,952,239.53


cuducos:
Just out of curiosity: isn't _'Taxi, toll and parking': work in progress_ merged already? You could link your notebooks back and forth ; )

Irio:
Great piece of work @fabiocorreacordeiro. Would like to suggest an extension for your research: although the limit of R$6000/month exists, deputies are allowed to pay for more than a single car - the requirement is the reason for using it: parliamentary activities. These cars may be used by the congressperson or advisors.

Given this information, it's also important to consider the number of cars receiving this amount of fuel and lubricants. As far as I know, they may be rented (the document will have the "Automotive vehicle renting or charter" `data.subquota_description` attribute) or belong to the congressperson himself (I'd bet that this hardly happens, given the existence of the previously mentioned subquota for rentals).

fabiocorreacordeiro:
Just to confirm, each line in the reimbursement database refers to only one receipt, is that correct? So, for the receipt to have the value above of the stipulated max cost we have to consider the hypothesis that the several cars were refuel at the same time.
Another question, the 'issue_date' is the date when the receipt was issued by the supplier or the reimbursement was asked by the congressman?

cuducos:
> each line in the reimbursement database refers to only one receipt, is that correct?

Yes.

> for the receipt to have the value above of the stipulated max cost we have to consider the hypothesis that the several cars were refuel at the same time.

At the some month (the subquota limit refers to the full month, so the congressperson can agree with the gas station to pay monthly, for example).

> the `issue_date` is the date when the receipt was issued by the supplier or the reimbursement was asked by the congressman?

When the receipt was issued: https://jarbas.serenatadeamor.org/static/ceap-datasets.html

fabiocorreacordeiro:
Ok,
I will analyze the subquota "Automotive vehicle renting or charter" to try cross the data.
Thanks

fabiocorreacordeiro:
@Irio, I am working again in the local transportation expenses and studying the CEAP legislation I found this paragraph:
```
§ 1º As despesas estabelecidas nos incisos I, VII e VIII poderão ser realizadas por assessores, assim entendidos os servidores efetivos, os ocupantes de cargos de natureza especial ou secretários parlamentares vinculados à Câmara dos Deputados. 

§ 1º The expenses established in items I, VII and VIII may be performed by advisors, understood as the effective servants, occupants of special positions or parliamentary secretaries attached to the Chamber of Deputies.
```
The "Fuels and Lubricants" were in item IX, so we should admit that only the congressman could use this kind of expenses and he could use only one car each time.

cuducos:
Hi @fabiocorreacordeiro,

Sorry I edited your comment — it was only for formatting purposes (making it easier to read without scrolling horizontally).

What you're doing is really valuable — we have [an open issue](https://github.com/datasciencebr/rosie/issues/3) about this research you started. This is great. I'll add more comments on #215 you opened ; )

PR #204, opened 2017-03-19
Refactor sex related places

henriquebraga:
*Fix docstring inconsistences;
*Passes empty str ('') at `company_name`, `company_trade_name` instead of `None` value;

### DIFF
--- a/src/fetch_sex_places.py
+++ b/src/fetch_sex_places.py
@@ -172,9 +172,9 @@ class SexPlacesNearBy:
                 'latitude': latitude,
                 'longitude': longitude,
                 'distance': distance.meters,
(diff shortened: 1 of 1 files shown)


src/fetch_sex_places.py
```diff
@@ -173,8 +173,8 @@ def parse(self, keyword, content):
                 'longitude': longitude,
                 'distance': distance.meters,
                 'cnpj': self.company.get('cnpj'),
```
cuducos: Any reason to leave `cnpj` behind?

henriquebraga: No, I haven't noticed :horse: it . Fixing it (thank you very much!).

cuducos:
Tks ; )

Issue #196, opened 2017-03-01
Find or create a database of all politicians


Irio:
As mentioned [in articles](https://datasciencebr.com/expandindo-a-serenata-de-amor-para-outras-esferas-48bed204d36d), although Serenata de Amor is currently focused in the Chamber of Deputies, the same models could be adapted for other Brazilian quotas, like Senate, Legislative Chambers, City Halls... Having a database of whole Politicians life - which is public information - can be very useful for exploration and Supervised Machine Learning. Related hypotheses are many: is a suspicious expense made in a place where the politician has a lot of influence? Is the politician in the first term or we could infer a significant influence in the Congress?

Example:

| Name | Full Name | Position | Place | Inauguration | Ending |
|------|-----------|----------|-------|-------|-----|
| Fernando Collor | Fernando Affonso Collor de Mello | Mayor | Maceió - AL | 1979-01-01 | 1983-01-01 |
| Fernando Collor | Fernando Affonso Collor de Mello | Federal Deputy | AL | 1983-02-01 | 1987-02-01 |
| Fernando Collor | Fernando Affonso Collor de Mello | State Governor | AL | 1987-03-15 | 1989-05-14 |
| Fernando Collor | Fernando Affonso Collor de Mello | President | Brazil | 1990-03-15 | 1992-12-29 |
| Fernando Collor | Fernando Affonso Collor de Mello | Senator | AL | 2007-02-01 |  |

May be appealing to scrape Wikipedia for that, but I wish we could use [TSE](http://www.tse.jus.br/) or other governmental sources. Candidacies is also a very important dataset to be collected next.

vitorbaptista:
http://everypolitician.org/ aims to be a database for all politicians across the globe, and could be an interesting source/partner. For instance, this is the data they have for Brazilian politicians: http://everypolitician.org/brazil/.

allantorres:
Great data but they don't have the dates but is a start, 
TSE have all data we need : http://www.tse.jus.br/eleicoes/estatisticas/repositorio-de-dados-eleitorais
But we are going to need to clean up the data ...

ppKrauss:
[...]

PR #208, opened 2017-03-31
created fetch_tse_data

rafonseca:
Hello guys.
This is the script that fetches data from TSE website in order to create a list of brazilian politicians. I have also a small example notebook that uses this data.
As the script will be code reviewed, I thought the notebook version would be appreciated. So this is a direct export from a notebook. Is it ok?


jtemporal:
Hi @rafonseca awesome contribution! I would recommend a few things here in order to start reviewing this PR =)

The `.py` generated from the notebook is awesome but it should be in the `develop/` directory alongside with the `.ipynb` file and the `.html` file. You can checkout [here](https://github.com/datasciencebr/serenata-de-amor/blob/master/CONTRIBUTING.md#best-practices) how to configure your Jupyter Notebook to generate all of these files while saving (if you haven't done so yet).

Also I would like to point out that inside the `src/` only goes the clean code, you can checkout other python scripts there if want ;)

So tl;dr for this PR. You'll need to do the following:
- [ ] move the `.py` to the `develop/` subdirectory
- [ ] commit both `.html` and `.ipynb` files in the `develop/` subdirectory
- [ ] create the clean `.py` script inside the `src/` subdirectory

rafonseca:
Ok, I've got it.
so, should I close this PR and create another one?

jtemporal:
No! you can do your changes here ;)
That's the workflow: you send things, we suggest some changes, you change it and it goes under review again. You are on the right track, great job!

cuducos:
Awesome contribution @rafonseca, awesome feedback @jtemporal!

No need to close the PR. You can work on changes on this branch and new commits are automagically added to this PR when pushed to GitHub ; )

rafonseca:
It is indeed magical

develop/2016-08-08-irio-translate-dataset.html
```diff
@@ -8456,8 +8456,8 @@
 }
 /* Flexible box model classes */
 /* Taken from Alex Russell http://infrequently.org/2009/08/css-3-progress/ */
-/* This file is a compatibility layer.  It allows the usage of flexible box 
-model layouts across multiple browsers, including older browsers.  The newest,
+/* This file is a compatability layer.  It allows the usage of flexible box
```
guizero: `compatibility` estava certo.

develop/2016-08-08-irio-translate-dataset.html
```diff
@@ -8456,8 +8456,8 @@
 }
 /* Flexible box model classes */
 /* Taken from Alex Russell http://infrequently.org/2009/08/css-3-progress/ */
-/* This file is a compatibility layer.  It allows the usage of flexible box 
-model layouts across multiple browsers, including older browsers.  The newest,
+/* This file is a compatability layer.  It allows the usage of flexible box 
+model layouts accross multiple browsers, including older browsers.  The newest,
```
guizero: `across` estava certo.

develop/2016-08-08-irio-translate-dataset.html
```diff
@@ -11357,7 +11357,7 @@
  * Author: Jupyter Development Team
  */
 /** WARNING IF YOU ARE EDITTING THIS FILE, if this is a .css file, It has a lot
- * of chance of being generated from the ../less/[samename].less file, you can
+ * of chance of beeing generated from the ../less/[samename].less file, you can
```
guizero: `being` estava certo

jtemporal:
Hi again, thank you for contribution @rafonseca and thank you @guizero for the review.

Actually irio's notebook shouldn't change.
- first because they weren't changes due to incorrect logic or analysis or any improvement on that aspect;
- and second because we try not to change notebooks since ideally the have historical reasons for being that way, by historical reasons I mean: in some point in the future we will be able to get the same results as the ones in the notebook if we replicate the same conditions such as python version, packages version and data version.

@rafonseca would you mind changing it back? that would also correct @guizero comments ;)

rafonseca:
Hi @jtemporal  and @guizero,
sorry for the mess. Probably, I opened an older version of this notebook and accidentaly commited it. I will change it back as soon as possible. Actually, I need to figure out how to do this. I think I need help there, hehe..

cuducos:
> I think I need help there, hehe...

Quick and dirty way: find Irio's notebook here on GitHub, click _Raw_, download and replace your local copy with it. Add a commit saying something like _Reverting Irio's notebook original version_. 

Classy way: play with `git revert` on the commits that changed these files, skip these files, `git commit` and `git rebase` to organize new commits if needed ; )

rafonseca:
Thanks for the tip @cuducos . Finally, I did a checkout on that file using the SHA. Not so dirty, not so classy. Inevitably, I will make further errors, so I will have the opportunity to try the classy way :)

src/fetch_tse_data.py
```diff
@@ -0,0 +1,254 @@
...
+import numpy as np
+import os
+import urllib
+import zipfile
+import glob
+
+
+
+FILENAME_PREFIX='consulta_cand_'
+TEMP_PATH = '../data/tse_temp'
```
cuducos: This might fail in Windows (there it's `\` not `/`). We recommend using `os.path.join` to make file and directory paths that would work in both worlds.

src/fetch_tse_data.py
```diff
@@ -0,0 +1,254 @@
...
+TSE_CANDIDATES_URL='http://agencia.tse.jus.br/estatistica/sead/odsele/consulta_cand/'
+OUTPUT_DATASET_PATH = '../data/2017-03-31-tse-candidates.xz'
```
cuducos: Same as line 12.

src/fetch_tse_data.py
```diff
@@ -0,0 +1,254 @@
...
+os.makedirs(TEMP_PATH)
```
cuducos: We could use `tempfile` to manage temporary directories, couldn't we?

src/fetch_tse_data.py
```diff
@@ -0,0 +1,254 @@
...
+
+# setting year range from 2004 to 2016. this will be modified further to 'from 1994 to 2016'
```
cuducos: Why do we generate years from 2004 to 2016 and, later, transform it? Wouldn't it worth it to start from 1994 right now?

src/fetch_tse_data.py
```diff
@@ -0,0 +1,254 @@
...
+year_list=[str(year) for year in (range(2004,2017,2))]
```
cuducos: Can we use `date.today().year` instead of a hardcoded `2017`?

src/fetch_tse_data.py
```diff
@@ -0,0 +1,254 @@
...
+
+# Download files
+for year in year_list:
+    filename=FILENAME_PREFIX+year+'.zip'
```
cuducos: Using `format` is recommended when concatenating more than 2 strings: `'{}{}.zip'.format(FILENAME_PREFIX, year)`.

src/fetch_tse_data.py
```diff
@@ -0,0 +1,254 @@
...
+    file_url=TSE_CANDIDATES_URL+filename
+    output_file=os.path.join(TEMP_PATH,filename)
```
cuducos: Some Python best practices are to use spaces between the object being defined and its value, and also after comas (e.g. `output_file = os.path.join(TEMP_PATH, filename)` instead of `output_file=os.path.join(TEMP_PATH,filename)`). Check [PEP8](http://pep8.readthedocs.io/) or [`prospector`](https://pypi.python.org/pypi/prospector/) if you are interested in this hints on code quality ; )

src/fetch_tse_data.py
```diff
@@ -0,0 +1,254 @@
...
+    filename=FILENAME_PREFIX+year+'.zip'
+    filepath=os.path.join(TEMP_PATH,filename)
+    zip_ref = zipfile.ZipFile(filepath, 'r')
+    zip_ref.extractall(TEMP_PATH)
+    zip_ref.close()
+
+
+# ### Adding the headers
+# The following headers were extracted from LEIAME.pdf in consulta_cand_2016.zip.
+header_consulta_cand_till2010=[
```
cuducos: Considering this headers section:

- [ ] We don't need the `#` with no comments at the end of the lines
- [ ]  Also I don't think we need the headers you skip with the `#`
- [ ] And we need to translate the headers to English ([example](https://github.com/datasciencebr/serenata-toolbox/blob/master/serenata_toolbox/ceap_dataset.py#L51-L151))

src/fetch_tse_data.py
```diff
@@ -0,0 +1,254 @@
...
+        elif ('2012' in file_i):
+            cand_df_i=(pd.read_csv('./'+file_i,sep=';',header=None,dtype=str,names=header_consulta_cand_at2012,encoding='iso-8859-1'))
+        else:
+            cand_df_i=(pd.read_csv('./'+file_i,sep=';',header=None,dtype=str,names=header_consulta_cand_till2010,encoding='iso-8859-1'))
+        cand_df=cand_df.append(cand_df_i[sel_columns])
+
+cand_df.index=cand_df.reset_index().index # this index contains no useful information
+
+# Exporting data
+cand_df.to_csv(OUTPUT_DATASET_PATH,encoding='iso-8859-1',compression='xz',header=True,index=False)
```
cuducos: Can we use `utf-8`? (same for lines 240-244).

src/fetch_tse_data.py
```diff
@@ -0,0 +1,254 @@
...
+###Concatenate all files in one pandas dataframe
+cand_df=pd.DataFrame()
+for year in year_list:
+    filesname=FILENAME_PREFIX+year+'*.txt'
+    filespath=os.path.join(TEMP_PATH,filesname)
+    files_of_the_year= sorted(glob.glob(filespath))
+    for file_i in files_of_the_year:
+        # the following cases do not take into account next elections. hopefully, TSE will add headers to the files
+        if ('2014' in file_i) or ('2016' in file_i):
+            cand_df_i=(pd.read_csv('./'+file_i,sep=';',header=None,dtype=str,names=header_consulta_cand_from2014,encoding='iso-8859-1'))
```
cuducos: Is it recommended to use `np.str` instead of `str`, @jtemporal?

rafonseca:
Hi @cuducos. Thanks for the feedback. 
I completely agree that the fetch notebook should not be included. I thought it could be useful for the reviewing process. But for the data, it is still not clear to me what dataset we want to keep:
-This script merges some data from TSE selecting some columns. This is dataset number 1 and we have it.
-The notebook 'use-tse-data' use dataset number 1 and produces dataset number 2, which is not saved in data/ (hell, it is a notebook!).
dataset2 is a nice readable list of all elected politicians as requested in the issue #196, but dataset1 has indeed more information...
Again, thanks for the review. I will work on the code soon.

src/fetch_tse_data.py
rafonseca: Well, the TSE informs that the data from 1994 to 2002 is not consistent, and they are working on this. I can try to get those data, but ...

src/fetch_tse_data.py
rafonseca: Yes, we can. It is more elegant, but the headers are hardcoded and there is a high probability that next election dataset will have different headers from all these...

src/fetch_tse_data.py
rafonseca: Very reasonable!

src/fetch_tse_data.py
rafonseca: Probably yes for exporting and no for importing. I will check again

src/fetch_tse_data.py
cuducos: Wow… no, no… now I got it. Just leave code as it is, but maybe clarify it in comments…

src/fetch_tse_data.py
cuducos: Damn… that's not good news. That's something that might be helpful to have documented (in comments) to future-me and future-you ; )

src/fetch_tse_data.py
cuducos: Yep, sure thing. My bad. The idea as to use UTF-8 for exporting, sorry ; )

cuducos:
Hum… now I got it.

I tend to agree that creating only the first dataset, and the notebook that demonstrated how to get from it to a list of elected politicians is good enough. 

@jtemporal and @Irio — you're more familiar than I am with data science — what do you think about it?

src/fetch_tse_data.py
jtemporal: Yes, indeed it is, in a few words: `<numpy str objects>` are different from `<python str objects>`. To keep the pattern used in the whole project we should stick to using `np.str` =)

jtemporal:
I personally would go to having the first dataset on the S3 and keep the notebook that shows " how to get from it to a list of elected politicians".
We usually have this: A large, structured, translated and ready to work dataset from which we can retrieve information and/or intelligence by filtering or applying some coding on top of it.
Maybe @Irio has some other point of view to share with us.

Irio:
@cuducos There's [a precedent](https://github.com/datasciencebr/serenata-de-amor/blob/fe9c5602c5ff496f8197777274b3c576faee1472/develop/2016-08-08-irio-translate-dataset.ipynb) for merging a notebook directly and only related to a `src` script, so I'm ok with keeping doing so.

cuducos:
@Irio AFAIK the notebook you mention **does not** write to `data/` — it's a study about how to implement a script that keeps the data directory intact. If @rafonseca's notebook doesn't write new data I'm fine.

Irio:
@rafonseca Do you need help for making these requested changes?

src/fetch_tse_data.py
rafonseca: Actually, I used "unicode" in python 2.7. In python 3, the standard str replaced unicode. We can use np.str as long as it deals with unicode characters. Does it?

rafonseca:
Hi @Irio.
Indeed, I have been quite busy last weeks and I will probably be busy for the next two weeks. Shame on me. I will try to find some gap to work on this.
I will certainly need help for the headers translation.
There is one last thing which is not clear for me. As you have noticed, I created the script from the notebook and there is important information on the notebook that will help to maintain this script in the future, like those infos @cuducos asked to comment somewhere. So, we are going to keep also the original notebook or not? Of course, we would remove the file writing in '../data'

rafonseca:
Hello there,
I did almost all modifications @cuducos  suggested. It is still missing the translation. Finally, I removed the scratch notebook and moved the important information to the script comments. I will try to answer more quickly for the next reviews. This PR is becoming long, hehe.
Thanks for the suggestions again

src/fetch_tse_data.py
```diff
@@ -0,0 +1,225 @@
...
+import urllib
+import zipfile
+import glob
+
+from tempfile import mkdtemp
+TEMP_PATH= mkdtemp()
+
+FILENAME_PREFIX= 'consulta_cand_'
+TSE_CANDIDATES_URL= 'http://agencia.tse.jus.br/estatistica/sead/odsele/consulta_cand/'
+OUTPUT_DATASET_PATH= os.path.join(os.pardir,'data','2017-03-31-tse-candidates.xz')
```
cuducos: It's not a good idea to have this date hardcoded here. Something like this might be useful:

```console
>>> from datetime import date
>>> today = date.today()
>>> today.strftime('%Y-%m-%d')
'2017-05-10'
```

cuducos:
>  It is still missing the translation.

If the last link got lost, [this might me helpful](https://github.com/datasciencebr/serenata-toolbox/blob/master/serenata_toolbox/chamber_of_deputies/chamber_of_deputies_dataset.py#L51-L153) ; )

> I will try to answer more quickly for the next reviews.

No worries — worst case scenario if we need this data ASAP we merge your PR to a secondary branch and work on some tweaks before bringing your code to the master branch…

> This PR is becoming long

No problem at all! This is valuable data <3

cuducos:
🎉 Many thanks @rafonseca ; )

rafonseca:
yuhuu!!
Thank you all. Hope it is the first PR of many!

PR #211, opened 2017-04-03
Make install instructions friendly for Windows users

cuducos:
Basically offer an alternative to nix OS `./` for execution and `source` for… well… sourcing.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -38,6 +38,13 @@ $ ./setup
 
 The `activate serenata_de_amor` command must be run every time you enter the project folder to start working.
 
+**If you use Windows** skip the last two commands, run _Command Prompt_ as _Administrator_ and use:
+
+```console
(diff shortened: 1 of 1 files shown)


rodolfo-viana:
@cuducos, @jtemporal, everyone related,

There is something else.

Windows users --especially regular ones, not developers or heavy users-- are not used to using Command Prompt. So it would solve some minor trouble if de description was something like this:

If you use Windows, run Command Prompt as Administrator and insert these commands instead:

$ activate serenata_de_amor
$ python setup

That is the only thing I remember was different from the original instructions.

cuducos:
Many thanks for the heads up @rodolfo-viana — @jtemporal anything else or RTM?

jtemporal:
To me it looks pretty good ;) 

nice job @cuducos and @rodolfo-viana thanks again for the feedback \o/

Issue #217, opened 2017-04-26
Update beautifulsoup4 version


jtemporal:
Conda requirements is trying to install beautifulsoup4 on version 4.4.1 and that is giving out an error while running the setup:

![photo_2017-04-26_14-09-53](https://cloud.githubusercontent.com/assets/6595551/25447881/30e3551e-2a8d-11e7-809d-92dab4b9a53f.jpg)
right now an appropriate version would be 4.5.3

jtemporal:
thanks to @sergiomario this one can be closed!

PR #220, opened 2017-04-29
Fix beautifulsoup4 version on requirements

sergiomario:
Fix the #217 issue.


cuducos:
🎉 Many tanks _môquiridu_

PR #218, opened 2017-04-27
Companies partners list in companies dataset - Issue #16

marcusrehm:
Just remembering. we can run the script as `python ./src/fetch_cnpj_info.py ./data/2016-11-19-current-year.xz -p 177.67.84.135:8080 177.67.82.80:8080 177.67.82.80:3128 179.185.54.114:8080 -t 20` where `-p` is for a proxies list and `-t` is for number of threads. Besides that we can pass -v for verbosity level, so with -v DEBUG we can see proxies errors and other return messages.

The script is doing the translate and transform logic that was in the clean script, so no need anymore for that one.

I'm finishing the dataset download, so when it's done I post the link ok? There are almost 65k companies info.

PR #221, opened 2017-04-29
Allow src/clean_cnpj_info_dataset.py scripts to receive data files as cli args

martini97:
As specified in [#167](https://github.com/datasciencebr/serenata-de-amor/issues/167), some scripts in src/ didn't accepted command line arguments, this adds an option to pass file arguments to the src/clean_cnpj_info_dataset.py. It treats arguments in the following way: if you pass two unnamed args to it, the first will be treated as the cnpj_cpf.xz file, and the second as the companies.xz file, you can also pass named arguments, using the --cnpj and --companies flags.

### DIFF
--- a/src/clean_cnpj_info_dataset.py
+++ b/src/clean_cnpj_info_dataset.py
@@ -16,10 +16,10 @@ parser.add_option("--companies", dest="companies", metavar="FILE",
 (options, args) = parser.parse_args()
 
 if options['cnpj_cpf']:
-    CNPJ_CPF_CSV_PATH = os.path.join('data', 'cnpj-info.xz')
+    CNPJ_CPF_CSV_PATH = options['cnpj_cpf']
 
 if options['companies']:
-    COMPANIES_CSV_PATH = os.path.join('data', 'companies.xz')
+    COMPANIES_CSV_PATH = options['companies']
 
 no_named_args = (not options['cnpj_cpf']) and (not options['companies'])
 
--- a/src/clean_cnpj_info_dataset.py
+++ b/src/clean_cnpj_info_dataset.py
@@ -3,6 +3,35 @@ import numpy as np
 import os
 import pandas as pd
 
+from optparse import OptionParser
+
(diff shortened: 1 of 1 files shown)


cuducos:
Many thanks for that, @martini97 — that's great news!

On other news I'd like to invite @marcusrehm: as you (Marcus) is working on a big refactor of this script, I thought that maybe we could merge this PR to #218 and then to `master` when it's time — do you think it's a good idea? Let us know ; )

@martini97: if that's the case would you mind sending the PR to @marcusrehm fork and branch? Your code will be merged here once we finish #218.

marcusrehm:
Hi guys!

After PR #218, the `clean_cnpj_info_dataset.py` will no longer exit :(  Its clean logic was moved to `fetch_cnpj_info.py`.

The `fetch_cnpj_info.py` script can receive as many positional arguments holding file paths as read source to extract CNPJ, but these files need to be mapped as commented [here](https://github.com/datasciencebr/serenata-de-amor/issues/167#issuecomment-298142624).

@martini97 you can see how it is now in PR #218 . If you have any comments we can adjust it.

martini97:
@marcusrehm What do you think about using a config file to map these files? If there are other places that use that info I think that this could be interesting

cuducos:
> What do you think about using a config file to map these files? If there are other places that use that info I think that this could be interesting

In deed ; )

cuducos:
Closed in favor of #218

marcusrehm:
Good idea @martini97 ! Think it would be nice in deed. Do you wanna do it forking my repo/branch or want me to do that?

martini97:
I would like to that, if It is ok with you

marcusrehm:
Cool @martini97 ! I will wait you do it and send me a PR at my fork, then we can merge it here after.

What do you think @cuducos ?!

martini97:
Great, I will start it tomorrow if that's ok

marcusrehm:
No problem! The branch is companies-partner-list ok?

martini97:
Ok

PR #223, opened 2017-04-29
Added some links to contributing guide, fixed setup (and changed it to a bash script)

### LINKED ISSUE #222: src/fetch_datasets.py integrated into toolbox, but still mentioned in docs
The contribution guide still points to `src/fetch_datasets.py`, which was removed in 1604567f55c782874e9ee98f979f98cd3ab76d60. I'll be opening a pull request soon recreating the file, now containing just these two lines:

```python
from serenata_toolbox.datasets import fetch_latest_backup
fetch_latest_backup('data/')
```

This way, the `setup` file can be changed to a bash script, only invoking python to run this new file on the last lines.

cauebs:
Closes #222

### DIFF
--- a/setup
+++ b/setup
@@ -12,4 +12,4 @@ fi
 
 echo "Downloading datasets (this might take several minutes depending on your internet connection)"
 
-python3 ./src/fetch_datasets.py
+python ./src/fetch_datasets.py
--- a/setup
+++ b/setup
@@ -1,21 +1,15 @@
-#!/usr/bin/env python3
+#!/usr/bin/bash
 
-import pip
-from shutil import copyfile, which
-import subprocess
+cp config.ini.example config.ini
 
-copyfile('config.ini.example', 'config.ini')
+pip install --upgrade -r requirements.txt
+conda pip config --add channels conda-forge
+conda install --yes --file conda_requirements.txt
 
-pip.main(['install', '--upgrade', '-r', 'requirements.txt'])
-subprocess.call(['conda', 'config', '--add', 'channels', 'conda-forge'])
-subprocess.call(['conda', 'install', '--yes', '--file', 'conda_requirements.txt'])
+if ! which unzip 1>/dev/null 2>/dev/null; then
+    echo "The following executables are needed and were not found: unzip"
+fi
 
-if which('unzip') == None:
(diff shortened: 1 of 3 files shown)


setup
```diff
@@ -1,21 +1,15 @@
...
-    print('The following executables are needed and were not found: unzip')
+echo "Downloading datasets (this might take several minutes depending on your internet connection)"
 
-print('Downloading datasets (this might take several minutes depending on your internet connection)')
-
-
-# This is imported here because serenata_toolbox is only installed when pip.main is called (line 9)
-from serenata_toolbox.datasets import fetch_latest_backup
-fetch_latest_backup('data/')
+python3 ./src/fetch_datasets.py
```
cuducos: As the default Python for the environment is Python 3, I think we should use `python` instead of `python3`. What do you think?

cauebs: I usually prefer to be explicit (because of weird distros like ubuntu), but in this case it shouldn't matter. Done!

setup
```diff
@@ -1,21 +1,15 @@
-#!/usr/bin/env python3
+#!/usr/bin/bash
```
cuducos: It doesn't seem to work here (macOS).

<details>

```console
$ ./setup
zsh: ./setup: bad interpreter: /usr/bin/bash: no such file or directory
$ which bash
/bin/bash
```

</details>


This leads me to another question: what about Windows users?

Maybe it's time to get rid of `setup` and replace it by a few command line instructions — so we can be more specific in addressing multiple platforms.

jtemporal: > This leads me to another question: what about Windows users?

they do not have bash, it usually is cmd

> Maybe it's time to get rid of setup and replace it by a few command line instructions — so we can be more specific in addressing multiple platforms.

Couldn't agree more!

cuducos:
@cauebs Considering [the last comments](https://github.com/datasciencebr/serenata-de-amor/pull/223#pullrequestreview-35783758) I think it is totally up to you:

* You can go ahead, `git rm setup` and add instructions to the `CONTRIBUTING.md` 
* Or just revert changes to `setup` and we handle that later

Besides that, thanks for the improvement in the documentation and to the port to the script that download the datasets ; )

cuducos:
Closed by inactivity

Issue #224, opened 2017-04-29
Using camara.leg.br API


caiopo:
Hi,

@luipillmann and @cauebs found out that camara.leg.br has an [API](http://www2.camara.leg.br/transparencia/dados-abertos/dados-abertos-legislativo/webservices).

There are some scripts in `src/` that scrape the HTML for data available on the endpoint [obterDetalhesDeputado](http://www2.camara.leg.br/transparencia/dados-abertos/dados-abertos-legislativo/webservices/deputados/obterDetalhesDeputado).

cuducos:
Hi @caiopo — can you me more specific on this issue? What is exactly the idea/proposal? Refactor some scripts? Which ones? Add new ones?Cheers ; )

caiopo:
`src/fetch_civil_names.py` is one of the scripts that could be refactored, but all the data it fetches is also fetched by `fetch_congressperson_details.py` (#225), so it could be deleted (if we assure it won't break anything).

cuducos:
I'm not sure about deleting it: if someone want to replicate some analysis made with this script having it might be useful… or not. What do you think @irio? Can we delete a script that was replaced by another one that fetches more stuff? The first one fetched only civil names, the one from #225 fetches civil names, gender and more.

caiopo:
> if someone want to replicate some analysis made with this

Can't we edit the notebooks to use the new script/dataset?

cuducos:
Sure, but then I would advise you to also rename the script (as fetch details might make me sense than fresh civil name after editing it)… that is to say: it will end up in the same situation…

cuducos:
I think collecting personal data on politician is still an issue. I would leave this because collecting this data might be useful to build something together with #119, for example.

PR #225, opened 2017-04-29
Add script to fetch congresspeople details

caiopo:
I talked to @joniarroba about useful data to create analysis that is available at Camara's API (e.g. age) but isn't fetched by the scripts in `src/`.

So this PR adds a new script that fetches and creates a `pandas.DataFrame` with `birth_date`, `civil_name`, `congressperson_id`, `gender`.

This script is "forked" from `src/fetch_civil_names.py`, but uses Camara's API (as noted on #224).


cuducos:
IMHO contributions in that sense should enhance or completely replace `src/fetch_civil_names.py` — and not add another script that generates (partially) duplicated data. Also we need to document this script in `CONTRIBUTING.md`.

TBH with this PR I believe we don't need `src/fetch_civil_names.py` anymore.

So what I suggest is: 

- [ ] Document this new script in `CONTRIBUTING.md`
- [ ] Mark `src/fetch_civil_names.py` as _deprecated_ in `CONTRIBUTING.md`
- [ ] Generate a dataset and share the link with us so we can add to the `serenata_toolbox` list of datasets (the list that is downloaded in the `setup` process)

What do you think about it?

caiopo:
>  Document this new script in CONTRIBUTING.md
> Mark src/fetch_civil_names.py as deprecated in CONTRIBUTING.md

Done (9b2331a)

>  Generate a dataset and share the link with us so we can add to the serenata_toolbox list of datasets (the list that is downloaded in the setup process)

https://we.tl/yrnPxCmQFA

The next step is probably refactoring anything that depends on `src/fetch_civil_names.py` to use `src/fetch_congressperson_details.py`, right?

jtemporal:
> The next step is probably refactoring anything that depends on src/fetch_civil_names.py to use src/fetch_congressperson_details.py, right?

That's about right, AFAIK there isn't anything depending on it... But if there is, yes, a refactor is in order.

cuducos:
I'm sorry @caiopo — it took a while for me to get back to this PR and the WeTransfer link is not working anymore. Can you re-upload it?

Besides that, do you confirm no refactor is needed as in:

> The next step is probably refactoring anything that depends on `src/fetch_civil_names.py` to use `src/fetch_congressperson_details.py`, right?

Next I think this is good to merge ; ) Many  many thanks

caiopo:
Hi! :smile: 

According to a `grep -ri "civil-names" develop/` (searches for `civil-names` in files inside `develop/`), no files depend on that dataset, so I deleted it on d7aaa824f133e89f0eee5c0d81a4bc2d1709bc71.

Here's the new WeTransfer link: https://we.tl/9WAO5WVpUW

I also fixed the conflicts so the branches can be automatically merged. :)

cuducos:
Many thanks, @caiopo 🎉

Issue #38, opened 2016-09-06
Exploratory Analysis on companies with the same name


Irio:
When collecting information about companies and working on other analyses, I found a large number of distinct companies with the same name (but different CNPJ's). Want to know more about their existence through an exploratory analysis, and possibly get more insights for new irregularities.

Here's a short list of company names with multiple CNPJ's in the quota:

```
AUTO POSTO CASCAO LTDA
IRMAOS BRETAS , FILHOS E CIA LTDA
FLECHA S.A. TURISMO COMERCIO E INDUSTRIA
POSTO AUTOMAN LTDA
ALLPARK EMPREENDIMENTOS, PARTICIPACOES E SERVICOS S.A.
RGB RESTAURANTES LTDA.
CANAA COMBUSTIVEIS PARA VEICULOS LTDA
MARIETTA COMERCIO DE ALIMENTOS LTDA
```

cuducos:
Having the name of the partners would be a good start. Issue #16 is crucial, but according to the discussion it's one of the more complex so far : /

PR #226, opened 2017-05-01
Adds notebook with intro to reimbursements dataset

luipillmann:
I thought it would be nice to have a quick overview on what these data can tell us. Here I suggest some initial questions a first-time reader/contributor could benefit from. Please take a look and check if it's worth it having it here 😄


develop/2017-05-01-luipillmann-intro-to-reimbursements.ipynb
```diff
@@ -0,0 +1,1715 @@
...
+   "metadata": {},
+   "source": [
+    "# Intro to reimbursements: overview with visualization\n",
+    "\n",
+    "This notebook provides an overview of the `2017-03-15-reimbursements.xz` dataset, which contains broad data regarding CEAP usage in all terms since 2009. \n",
+    "\n",
+    "It aims to provide an example of basic analyses and visualization by exploring topics such as:\n",
+    "\n",
+    "- Evolution of avarege monthly spending along the years\n",
+    "- Avarege monthly spending per congressperson along the years\n",
```
jtemporal: Typo found here: Average ;)

luipillmann:
Oops. Thanks @jtemporal, it's fixed.

luipillmann:
I noticed some misleading results when evaluating most hired suppliers. It is due to multiple 'aliases' suppliers have in the dataset, which I as not aware of (e.g. there are **hundreds** of variations like 'TELEFONICA S.A.',  'TELEFONICA S.A. - VIVO' for the same supplier).

Now I'm just grouping by `cnpj_cpf` and associating one of the aliases to this value.

jtemporal:
> Now I'm just grouping by cnpj_cpf and associating one of the aliases to this value.

nice!

I'll take a better look at this notebook today, but at first glance it is good to merge ;) Nice work @luipillmann

cuducos:
> II noticed some misleading results when evaluating most hired suppliers. It is due to multiple 'aliases' suppliers have in the dataset, which I was not aware of (e.g. there are **hundreds** of variations like _TELEFONICA S.A._,  _TELEFONICA (11) 3862.5534_, _`Telefônica Brasil S.A. - VIVO_ for the same supplier).

Maybe this data is typed by the user (in theory the congressperson) so we might have some differences. Anyway this might be of interest: #38

> Now I'm just grouping by `cnpj_cpf` and associating one of the aliases to this value.

Great strategy!

One thing I missed in your analysis made me a bit annoyed: it's too naïve to consider the total amount spend in different years without taking into account two questions.

1. has the quota increased/decreased in these years?
2. what about inflation in this period?

If you can bring in this data (hardcoded in your notebook, linking the source) it would be awesome ; )

luipillmann:
Nice, @cuducos!

Indeed, these are crucial considerations I've missed. I'll take the time to include those points and push a new version then.

Regarding the CNPJ issue, I agree it might be due to mistyped information. Although #38 relates to that (1 name, multiple CNPJs), what I pointed out was the opposite (1 CNPJ, multiple names). I think in both cases there is room for work:
- Analyze whether congressmen consistently registered the right CNPJ for the name (and vice-versa);
- Create a lookup dataset with CNPJ and its most recent/probable associated name.

luipillmann:
After a quite long time, I managed get back to work and fix the issue regarding inflation and yearly CEAP limits. I decided to plot all aspects in a single chart (avg. spending, CEAP limits and IPCA indexes) and, indeed, these data might explain the increase in congressmen spending.

I've also added some brief commentaries to main charts aiming to enhance legibility and suggesting further analysis in some cases (e.g. when comparing spending by states and parties). 

Please take a look and let me know if you have any other suggestions to make this notebook more useful.

cuducos:
Wow! many thanks, @luipillmann!

LGTM, @jtemporal — leaving it to you as you've said you would go through this notebook throughly  one of this days ; )

jtemporal:
It took me while to get back to this, but this looks great! Great work @luipillmann ;)

PR #227, opened 2017-05-02
Senate dataset first analysis notebook.

anaschwendler:
As we say in [this article](https://medium.com/data-science-brigade/expandindo-a-serenata-de-amor-para-outras-esferas-48bed204d36d), we are expanding our project to other datasets, the first one will be the Senate Quota dataset.

The datasets may be still added here later.
We will start by adding them to the `serenata-toolbox`, then we can use it here.

TODO:
- [ ] Create a script in the `serenata-toolbox` repository.
- [ ] Add the new datasets to the amazon repository.
- [ ] Add here the datasets to be explored.


cuducos:
LGTM ; )

PR #239, opened 2017-05-19
Update docker image and fix issues with conda packages installation

fgrehm:
Not sure if this is going to break things for people running the env outside of Docker but this is what I've done to get up and running again after reformating my laptop.

Tks to @cabral for the tip on using the `bioconda` channel.

BTW, things look a lot nicer on this new jupyter release 😍 

screen shot 2017-05-19 at 10 59 34

/cc @cuducos

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -72,7 +72,7 @@ $ source /usr/local/var/pyenv/versions/anaconda3-4.1.1/bin/activate serenata_de_
 
 You can user [Docker](https://docs.docker.com/engine/installation/) and [Docker Compose](https://docs.docker.com/compose/install/) to have a working environment:
 
-1. Start the environment (it might take a while, the base image has 5.8GB and we also pull in lots of dependencies): `$ docker-compose up -d`.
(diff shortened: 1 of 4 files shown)


jtemporal:
I doesn't look likely to break anything for people that do not use Docker... I'll test it out later today to be sure ;)

>BTW, things look a lot nicer on this new jupyter release 😍

It really does look nicer, doesn't it?! It is a shame that GitHub do not render the tables like that though ¯\\\_(ツ)_/¯

cuducos:
As I comment to @cabral last week I think a more stable fix would be to to install `grequest` from `pip`, not from `conda` (after all it has nothing to do — directly — with data science, so `pip` migh handle it better). That way we don't depend on `bioconda` (we can remove that). Do you mind addressing it too in this PR?

cuducos:
LGTM. @jtemporal I saw you struggling with `gresquests` in #244 — any thoughts here? I'll let for you to merge, ok?

jtemporal:
>@jtemporal I saw you struggling with gresquests in #244 — any thoughts here? 

@cuducos Yes I was, but @rougeth came up with a nice way to solve the problem I was having =)

So merging this one =)

PR #240, opened 2017-05-19
Fix a translation typo

lipemorais:


anaschwendler:
Thanks for this contribuition, @lipemorais!

PR #243, opened 2017-05-22
Federal Senate invalid cnpj cpf analysis

anaschwendler:
@jtemporal worked on a notebook to see what we would need to adapt to introducing the Federal Senate dataset to work on its first classifier, that is our final study and now the Federal Senate is already working with `invalid_cpnj_cpf`.

🎉


anaschwendler:
For me it is already good to merge, but since I and @jtemporal worked on it, I would like if that some other person to do so :)

PR #244, opened 2017-05-25
Cities Transparency Portal

jtemporal:
This PR aims to find all the transparency portals URLs for Brazilian cities


jtemporal:
Things that need to be done on the script:

- Receber `data_path` e `cities_file` como argumento
- state.upper()
- return type(‘Response’, (object,), {‘status_code’: ‘None’})

src/grequests_transparency_portal_cities.py
```diff
@@ -0,0 +1,86 @@
...
+    if not response.status_code:
+        return 404
+    return response.status_code
+
+
+def main(data_path='/tmp/serenata-data', cities_file='2017-05-22-brazilian-cities.csv'):
+    Datasets(data_path).downloader.download(cities_file)
+    br_cities = pd.read_csv(os.path.join(data_path, cities_file))
+
+    br_cities['state'] = br_cities['state'].apply(lambda x: x.lower())
```
cuducos: Just out of curiosity (I guess)… I tend to find `str.lower` more pythonic than `lambda x: x.lower()` (but it works just the same) — I mean no need to change that, just chitchatting…

src/grequests_transparency_portal_cities.py
```diff
@@ -0,0 +1,86 @@
...
+
+    br_cities['normalized_name'] = br_cities['name'] \
+                                              .apply(lambda x: normalize_string(x))
```
cuducos: `lambda x: whatever(x)` is just the same as `whatever`, isn't it? Can we test only `….apply(normalize_string)`?

develop/2017-05-19-jtemporal-cities-transparency-portal.py
```diff
@@ -0,0 +1,316 @@
...
+
+
+# In[16]:
+
+brazilian_cities.head()
+
+
+# ## Getting all cities that are part of Transparency Portal
+# 
+# There are some cities that we already know that have a page with transparency and open data. The main object here is to find how many cities have that.
```
cuducos: object -> objective ?

src/grequests_transparency_portal_cities.py
anaschwendler: It is fine, it is a good point :)

src/grequests_transparency_portal_cities.py
anaschwendler: I think that here it maybe necessary, it is raising this error: `TypeError: normalize_string() missing 1 required positional argument: 'string'`

anaschwendler: image

develop/2017-05-19-jtemporal-cities-transparency-portal.py
anaschwendler: fixed!

src/grequests_transparency_portal_cities.py
cuducos: Use `normalize_string` (the reference to the function in the namespace) not `normalize_string()` (a call to the function without any parameter) ; )

anaschwendler:
@cuducos so we already finished refactoring the script and study notebook.

you can check it out for us, and if it is ok, so we think it is good to merge :)

:tada:

src/grequests_transparency_portal_cities.py
```diff
@@ -0,0 +1,93 @@
+import os
```
cuducos: This script (`src/grequests_transparency_portal_cities.py`) has to be documented in `CONTRIBUTING.md` before merging, ok?

src/grequests_transparency_portal_cities.py
```diff
@@ -0,0 +1,93 @@
...
+if __name__ == '__main__':
+    description = """
+    This script generates a CSV file containing all the Brazilian
+    cities and the URLs for their transparency portal if a transparency portal
+    exists
+    """
+
+    parser = ArgumentParser(description=description)
+    parser.add_argument(
+        '--cities_file', '-c', default='2017-05-22-brazilian-cities.csv',
```
cuducos: A best practice (AFAIK) is to avoid underscores; would you mind changing it to `--cities-file`? (TBH, `--cities` sounds good enough for me — it's up to you)

src/grequests_transparency_portal_cities.py
```diff
@@ -0,0 +1,93 @@
...
+        help=('A CSV file containing all Brazilian cities per state '
+              '(default: 2017-05-22-brazilian-cities.csv)')
```
cuducos: This dataset (`2017-05-22-brazilian-cities.csv`) needs to be documented in `CONTRIBUTING.md` before merging, ok?

jtemporal: Wouldn't it be better if it were downloaded with the toolbox? I noticed now that it isn't at the `latest_backup` files list. What do you think?

cuducos: Isn't that what's happening in lines 87-88?

src/grequests_transparency_portal_cities.py
jtemporal: fixed on c7019bd

src/grequests_transparency_portal_cities.py
jtemporal: yep

jtemporal: fixed on 7cdc68a

src/grequests_transparency_portal_cities.py
jtemporal: fixed on 	c43ee2a

PR #249, opened 2017-06-12
Adds analysis for article on taxi vs apps

jtemporal:
as @vilapedro requested and thanks to both him and @anaschwendler ❤️


deliver/2017-06-12-jtemporal-anaschwendler-taxi-uber-bsb-analysis.py
```diff
@@ -0,0 +1,148 @@
...
+
+
+# In[8]:
+
+dataset.shape
+
+
+# ## Further investigation
+# 
+# With that table we selected a few companies and apps for futher investigation:
```
cuducos: I think we could add the rational (even if subjective) for this set of companies.

jtemporal: This notebook will be used as basis for an article, fell it exists more for an "How did you get this data?" kind of question once the article is out than an exploratory analysis which requires proper explanations.

cuducos: So… should I wait for the article to be publish and linked in the notebook?

jtemporal: I wouldn't even link it, since Irio didn't in a previous notebook in `deliver/` but I'm totally fine with that ;)

cuducos: Ow… @Irio is such a troublemaker. I'll add him as a reviewer because as far as I could get the `deliver/` notebooks should be a bit different (his notebook included). But maybe it's just me.

cuducos:
@irio, I invited you here. In general I think your `deliver/` notebook — basically a dump of the code we used for the Medium post — is not what I expected from `CONTRIBUTING.md`.

Is `deliver/` what we described as `report/`?

**If so**

I think I merged your notebook as a one time stunt: I wanted it so bad to have the code you've wrote shared here. _But_ I was shortsighted: that quality of _deliver_ is a kind of precedent that does not fit IMHO what we have described [we expected](https://github.com/datasciencebr/serenata-de-amor/blob/master/CONTRIBUTING.md#best-practices) for these documents:

> This is where we write up the findings and results, here is where we put together different data, analysis and strategies to make a point, feel free to jump in.

> Meaningful title for the report (e.g. `Transport-allowances.ipynb`)
 
These write ups, I'm afraid I expected to be more focused on general public than developers and data scientists — however both notebooks (yours and the one from this PR) are mostly code IMHO.

Also I think these notebooks are supposed to be in pt_BR (aka accessible for a larger _general_ public) — but not sure, I might be wrong.

**If not**

What the heck is supposed to be in `deliver/`?

**Bonus question**

Are we missing a space for specific questions asked/developed by journalists? IMHO this fits better `develop/` than `deliver/`… what do you all think about it?

Irio:
@cuducos
> that quality of deliver is a kind of precedent that does not fit IMHO what we have described we expected for these documents
Agree with you on that. In one year of history, we never used `/deliver` as we wanted, but [Medium](http://medium.com/data-science-brigade) directly.

My proposal is to just move `/deliver`'s contents to `/develop` and delete the original folder with all of its references.

cuducos:
> `My proposal is to just move `/deliver`'s contents to `/develop` and delete the original folder with all of its references.`

What about the notebook in pt-BR? Delete or translate it? Or just leave it as it is?

jtemporal:
maybe move the one in pt-br and open up an issue to translate it later?

cuducos:
> maybe move the one in pt-br and open up an issue to translate it later?

I disagree because of precedence… if anyone sees the notebook and start one of their own in pt-BR it will be a second translation needed. The easiest way IMHO is to move the notebook form this PR to `develop` and keep the pt-BR in `deliver/`…

cuducos:
Closed as after merging Jarbas and Rosie we're no longer storing notebooks

Issue #261, opened 2017-07-20
Fetch TSE data scrapper is broken


jtemporal:
As I was helping @gabrielacaesar earlier today we found out that [`fetch_tse_data.py`](https://github.com/datasciencebr/serenata-de-amor/blob/master/src/fetch_tse_data.py) was broken as follows:

<img width="490" alt="screen shot 2017-07-19 at 10 04 37 pm" src="https://user-images.githubusercontent.com/6595551/28396046-2e6624c2-6ccf-11e7-8eaa-beac865071a9.png">

we also tested is simply because onde you use `.isoformat()` what you have is an string and can't have  something like `isoformat().year`.

PR #262, opened 2017-07-20
Fixing fetch tse data scrapper

jtemporal:
this is a fix for #261


jtemporal:
I would like to bring to attention that the use of `os.pardir` breaks this scrapper if you try to run it from the root of the project:
```console
$ python src/fetch_tse_data.py
```

screen shot 2017-07-19 at 10 20 57 pm
and with that in mind should we remove `os.pardir` from the method that creates the output file path?

PR #265, opened 2017-07-23
Setup CI

Irio:
Tests will ensure the existence of one, and only one, html/py file for every Jupyter Notebook located in `develop` folder. Forgetting a version, or adding duplicated files fill make the suite fail.


test/test_serenata_de_amor.py
```diff
@@ -0,0 +1,23 @@
+import glob
+import subprocess
```
jtemporal: this is imported but not used.

PR #270, opened 2017-07-29
Move notebooks to research folder

Irio:
With this getting merged, I can work on bringing other Serenata's repositories as submodules so people willing to use Docker can have everything more integrated.

---

*edit for clarification*

This is an initial pull request aiming to gather all official projects from Serenata de Amor - this one with research work, Jarbas, Rosie, serenata-toolbox and whistleblower - into a single Git repository. The idea is to use [Git Submodules](https://git-scm.com/book/en/v2/Git-Tools-Submodules) so we can `git clone` the existing repositories into folders inside the one hosted here in https://github.com/datasciencebr/serenata-de-amor.

**Example when research and rosie folders are present**
```
irio@Irios-MacBook-Pro ~/C/s/serenata-de-amor (irio-docker)> tree
.
├── CONTRIBUTING.md
├── LICENSE
├── README.md
├── config.ini.example
├── docker-compose.yml
├── docs
│   ├── CEAP.md
│   ├── CEAP_pt_BR.md
│   ├── companies-with-federal-sanctions-datasets.md
│   ├── federal-budget-agreements-datasets.md
│   └── receipts-ocr.md
├── logo.png
├── research
│   ├── Dockerfile
│   ├── data
│   ├── develop
│   │   ├── 2017-07-17-cuducos-consultancies.html
│   │   ├── 2017-07-17-cuducos-consultancies.ipynb
│   │   └── 2017-07-17-cuducos-consultancies.py
│   ├── jupyter_notebook_config.py
│   ├── requirements.txt
│   ├── src
│   │   ├── backup_data.py
│   └── test
├── rosie
│   ├── Dockerfile
│   ├── LICENSE
│   ├── README.md
│   ├── config.ini.example
│   ├── requirements.txt
│   ├── rosie
│   │   ├── chamber_of_deputies
│   │   ├── core
│   │   └── federal_senate
│   ├── rosie.py
│   └── setup
└── setup
```

This is an idea getting dust since ~January in this branch, so I'd like to propose it through a minimum viable pull request. Hopefully, it's a change that will make easier for users willing to use Docker, needing to dedicate less time on configuration, thus being able to see by themselves how every project is connected to others (today we rely mostly on documentation). Also, sharing databases or data folders across projects should be much easier to manage.


jtemporal:
I'm afraid that using git submodules might make it harder for people to contribute. Any thoughts on that?

silviodc:
So if i understood and @jtemporal  mentioned you will include Rosie as a submodule. Ok I guess it could be good and will allow many people to produce full approaches and classifiers to detect suspicious reimbursements. 

However be careful using it. IMHO when you point submodule’s MASTER HEAD, you get many problems, to merge and manage the project.

jtemporal:
>However be careful using it. IMHO when you point submodule’s MASTER HEAD, you get many problems, to merge and manage the project.

That's exactly what I worry about. Especially considering that we have a lot of "non-tech" contributors

Irio:
Some of the arguments in favor or against of this decision are reflected in facts, others only in personal preferences. I will try my best to explicitly state where I'm expressing my opinions.

When internet connection, local storage and a bit overhead in the performance is not a problem, I believe using Docker Compose and Git Submodules is the most efficient way of working with multiple interdependent projects, at least the way we have in Serenata. Internet connection and local storage are problems that I can work on solving, just don't want to introduce larger changes before getting this initial branch merged into master. The first can be solved triggering the creation of our own Dockerfile at [Docker Hub](https://hub.docker.com/) after every `Dockerfile` change via CI; the latter using [Alpine](https://hub.docker.com/_/alpine/)-based images).

The new installation instructions could be something like this:

---

***README.md***

Install [Docker](https://www.docker.com/) and [Docker Compose](https://docs.docker.com/compose/). Then run the following:

```console
$ git clone --recursive --jobs 8 git@github.com:datasciencebr/serenata-de-amor.git
$ docker-compose up
```

Done. After download and setup finishes, you have every Serenata de Amor project interconnected with each other. Jupyter Notebook is available in the port 8888.

---

How this interconnection may benefit the contributors:

* Changes in the Toolbox will automatically reflect (without the need of committing, manually-triggered install or package release) in the local environment, as in Rosie, Jarbas and Whistleblower. IMO It takes a good amount of patience to test, in other systems, a Toolbox change.
* Rosie may share a filesystem folder with Jarbas. In this year working with Serenata in a daily-basis, I gathered multiple `data` folders which I often have to remember to delete. A common problem: I want to download new reimbursements from the government, run Rosie and explore the changes in a notebook. Should I use `/tmp/serenata-data`, which will be deleted in the next restart, `research/data` or somewhere else? Depending on my mood, I put at `/tmp/serenata-data`, inside Rosie, Whistleblower or in the research project. IMO We should provide a better default since it's a common problem.
* Currently, Serenata is using 3 databases: Postgres (Jarbas), MongoDB (Whistleblower) and filesystem (Rosie and Research). When considered separate projects, these choices might sound good decisions. When seen as parts of a single project, we may prefer other alternatives. An in-memory database plus Postgres, for instance.
* Local setup and deployment. Yes, we could write a script to connect everything together and provision/deploy to a new instance with fewer commands, but I believe using a tool made to solve this problem is easier and better than doing it on our own. "Standing on the shoulders of giants"...

For these reasons mostly I believe this is the ideal method **for active contributors**.

**Thinking about people with metathesiophobia and novices** in the project (without good experience with Git and Docker), unwilling to follow a possibly larger set of instructions, to get all the projects running and not just a single one. What will change for them after merging this pull request?

```console
$ git clone git@github.com:datasciencebr/serenata-de-amor.git
$ ./setup
$ jupyter notebook
```

This will work almost as before. One single change: Jupyter will show notebooks not in `develop` folder, but in `research/develop`.

What happens after bringing Rosie as a submodule, to come in a following pull request? A `rosie` file will appear in the project root. The content is a hash. They may completely ignore it.

What happens after bringing Jarbas as a submodule? A `jarbas` file will appear in the project root. The content is a hash.

**Changes in the work of contributors benefitting from the changes**: they will have a single folder with all Serenata de Amor projects. For committing in Jarbas, for instance, they will `cd` into `/jarbas` folder, do their Git work as usual (`git checkout -b branch-name`, `git commit`) there, still inside Serenata's main folder. Why I believe we won't have (more than usual) issues with integration? When using Git Submodules, even when pointing to master branch, it is not automatically updated to the most recent commit in master. In Git itself, when referencing to `master`, it's just [a reference](https://git-scm.com/book/en/v2/Git-Internals-Git-References) to a given commit, which is what is stored where is supposed to have a Git Submodule.

**TL;DR** I don't believe it will negatively effect contributors. The ones not willing to change to Docker Compose will have a very small change in their work; the others, will benefit in a daily-basis. The effects that may be considered negative for the ones using Docker have workarounds to be considered in pull requests to come.

anaschwendler:
Hi everyone!

I'm testing this PR and I believe that reading the reasons why move the repo to submodules will improve the experience in contributing to the project, making it the easier to run. It is the first move to turn the project easier and more attractive to both novices and experienced contributors.
We can be using Docker for it, but for the ones that don't want to use it, there will be still an option to non-Docker users. Docker will be only facilitating things @jtemporal :)

By now, what I did to test this PR (and @irio's suggestion of refactor):
FIrst, I got sure that I have Docker, and it is working fine.
I cloned the repo: 
```console
git clone --recursive --jobs 8 git@github.com:datasciencebr/serenata-de-amor.git
cd serenata-de-amor
```
Then I checkout to @Irio's branch: #270 
```console
git checkout -b irio-docker origin/irio-docker
```
And finally run the Docker command to get the project up and running:
```
docker-compose up
```
It may take some time by now, but I think that it will be improved, from now on. And, we can see it working here:
image

The only thing that I suggest before merging it is to update the `README.md` file, and document the option for the ones who doesn't want to use Docker. 
Other thing that I notice is that may be needed to create an example of how running it with this changes.
And then LGTM ;)

anaschwendler:
LGTM ;)

PR #277, opened 2017-08-29
Fix paths for Windows users

cuducos:
The latest version of `setup` is not compatible with some Windows users because of hardcoded `/` in directory paths. This PR fixes that ; )

### DIFF
--- a/setup
+++ b/setup
@@ -4,10 +4,15 @@ import os
 import pip
 from shutil import copyfile, which
 
+
+RESEARCH_DIR = 'research'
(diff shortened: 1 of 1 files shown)


jtemporal:
Great! Thanks for that @cuducos I'm sure it will be much appreciated by our windows users ❤

How I tested this:
- On my local clone updated the current code base using: `$ git pull`
- Then changed to @cuducos branches to check the proposed changes with: `git checkout cuducos-fix-windows-paths`
- since the changes were on the setup file just did ran it by `$ ./setup` and since no errors or warnings were raised I merged this PR =)

Issue #276, opened 2017-08-23
(Docker) PermissionError: [Errno 13]


joaopcanario:
I don't know what happened but after last updates from master branch I'm seeing this error on logs of my container:

`PermissionError: [Errno 13] Permission denied: '/Home/jovyan/.jupyter/migrated'`

I started to noticed as well that Jupyter isn't saving the .html and .py files of my notebook anymore. Someone already saw this?

joaopcanario:
And here my container log:

```
[I 12:51:12.660 NotebookApp] Saving file at /work/develop/2017-08-20-joaopcanario-companies-analisys.ipynb
[W 12:51:12.661 NotebookApp] Saving untrusted notebook work/develop/2017-08-20-joaopcanario-companies-analisys.ipynb
Traceback (most recent call last):
  File "/opt/conda/bin/jupyter-nbconvert", line 6, in <module>
    sys.exit(nbconvert.nbconvertapp.main())
  File "/opt/conda/lib/python3.6/site-packages/jupyter_core/application.py", line 267, in launch_instance
    return super(JupyterApp, cls).launch_instance(argv=argv, **kwargs)
  File "/opt/conda/lib/python3.6/site-packages/traitlets/config/application.py", line 657, in launch_instance
    app.initialize(argv)
  File "<decorator-gen-7>", line 2, in initialize
  File "/opt/conda/lib/python3.6/site-packages/traitlets/config/application.py", line 87, in catch_config_error
    return method(app, *args, **kwargs)
  File "/opt/conda/lib/python3.6/site-packages/nbconvert/nbconvertapp.py", line 257, in initialize
    super(NbConvertApp, self).initialize(argv)
  File "<decorator-gen-6>", line 2, in initialize
  File "/opt/conda/lib/python3.6/site-packages/traitlets/config/application.py", line 87, in catch_config_error
    return method(app, *args, **kwargs)
  File "/opt/conda/lib/python3.6/site-packages/jupyter_core/application.py", line 243, in initialize
    self.migrate_config()
  File "/opt/conda/lib/python3.6/site-packages/jupyter_core/application.py", line 169, in migrate_config
    migrate()
  File "/opt/conda/lib/python3.6/site-packages/jupyter_core/migrate.py", line 241, in migrate
    with open(os.path.join(env['jupyter_config'], 'migrated'), 'w') as f:
PermissionError: [Errno 13] Permission denied: '/home/jovyan/.jupyter/migrated'
[...]

PR #279, opened 2017-09-11
#276 - Fixed error on editing Jupyter notebook

viniciusartur:
Bug:
  Given serenata-de-amor cloned from github, built by docker-compose and up by docker-compose
  When editing a Jupyter notebook from directory work/develop
  Then causes "PermissionError: [Errno 13] Permission denied: '/home/jovyan/.jupyter/migrated'"

Bug analysis:
  Directory /home/jovyan/.jupyter with owner root and jupyter process with owner jovyan

Root cause:
  COPY command of Docker creates files and directories with a UID and GID of 0, according to [documentation](https://docs.docker.com/engine/reference/builder/#copy) and pointed on this [thread](https://stackoverflow.com/questions/44766665/how-do-i-docker-copy-as-non-root).

Solution:
  Setting owner jovyan for directory /home/jovyan/.jupyter

### DIFF
--- a/research/Dockerfile
+++ b/research/Dockerfile
@@ -9,8 +9,9 @@ RUN apt-get update \
         libxml2-dev \
         libxslt-dev \
     && apt-get clean \
-    && rm -rf /var/lib/apt/lists/* \
-    && sudo -u jovyan mkdir /home/jovyan/.jupyter
+    && rm -rf /var/lib/apt/lists/*
+
+RUN sudo -u jovyan mkdir /home/jovyan/.jupyter
 
 USER jovyan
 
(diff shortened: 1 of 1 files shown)


research/Dockerfile
```diff
@@ -9,7 +9,8 @@ RUN apt-get update \
         libxml2-dev \
         libxslt-dev \
     && apt-get clean \
-    && rm -rf /var/lib/apt/lists/*
+    && rm -rf /var/lib/apt/lists/* \
+    && sudo -u jovyan mkdir /home/jovyan/.jupyter
```
cuducos: Would you mind adding `sudo -u jovyan mkdir /home/jovyan/.jupyter` as a single `RUN` step?

My point is that it doesn't have to do with the logic of the previous `RUN` (namely, with installing dependencies).

I think it makes sense to end up with a `RUN` step to install dependencies and another `RUN`step to fix permissions.

Does it make sense?

viniciusartur:
Requirements enlightened and code refactored!

Issue #283, opened 2017-10-03
Add a donate button to README


jtemporal:
**What is the problem?**
We don't have a donate button 😱

**How can this be addressed?**
Add a button donate button to the README file

It something like ![donate](https://img.shields.io/badge/donate-apoia.se-EB4A3B.svg) linked with [Serenata's Apoia.se Campaign](https://apoia.se/serenata). I generated that badge with [Shields.io](https://shields.io).

**Who could help with this issue?**
Anyone really

**Labels**
Already there 😉

**Important**
This is a [first-timers-only issue](http://www.firsttimersonly.com), it is supposed to help people get their feet wet. Keep that in mind when considering to open a pull request 😉

jtemporal:
Code for the badge mentioned: `![donate](https://img.shields.io/badge/donate-apoia.se-EB4A3B.svg)`

lemuellloren:
#285

cuducos:
IMHO this is not necessary in this repo. We already have a link in the `README.md` and actually we have no set of badges to put together with a badge as such. What do you think, @jtemporal?

maykillmore:
Can I take this up?

cuducos:
Closed by #284

PR #284, opened 2017-10-03
Add donate button

richardljs:
Adding a donate button to the README file, referencing https://github.com/datasciencebr/serenata-de-amor/issues/283.


richardljs:
Yes, that makes sense. Thanks for the suggestion!

PR #335, opened 2018-03-01
Configure Papertrail in production environments

Irio:


.env.example
```diff
@@ -0,0 +1 @@
+LOGGING_URL=
```
cuducos: Can we have this in [`contrib/.env.sample`](https://github.com/okfn-brasil/serenata-de-amor/blob/master/contrib/.env.sample) instead?

Irio: done

cuducos: Hum… actually I might have not expressed myself properly. Do we need two `.env.sample` files? One in root (added by this PR) **and**  the other one already included in `contrib/` directory?

IMHO either we:

* add the line `LOGGING_URL=` to the existing `contrib/.env.sample`;
* **or** we move the existing `contrib/.env.sample` to root and merge it to the `.env.sample`/`.env.example` included in this PR.

So we end up with _only one_ `.env.sample`. Does that make sense?

Irio: My bad. Can you check now, after the changes?

docker-compose.prod.yml
```diff
@@ -84,12 +87,13 @@ services:
...
       HTTPS_METHOD: ${HTTPS_METHOD:-redirect}
       VIRTUAL_PROTO: http
+    hostname: nginx
 
   logspout:
     image: gliderlabs/logspout:v3.2.4
     restart: always
     env_file:
-      - .env
+      - contrib/.env
```
cuducos: For production we're using `.env.prod` (in the root dir), [as in other containers in this same file](https://github.com/okfn-brasil/serenata-de-amor/blob/6b4cb7c8367a1436d38adc11029e22cfb8318f08/docker-compose.prod.yml#L38-L39).

cuducos:
I tried to run it with production settings and even adding the proper `LOGGING_URL` in the `.env.prod` I got an warning:

```
WARNING: The LOGGING_URL variable is not set. Defaulting to a blank string.
```

Any ideas about what I might be doing wrong?

Issue #338, opened 2018-03-07
Three undefined names


cclauss:
Flake8 testing of https://github.com/okfn-brasil/serenata-de-amor on Python 3.6.3

$ __flake8 . --count --select=E901,E999,F821,F822,F823 --show-source --statistics__
```
./research/src/fetch_inbox.py:21:18: F821 undefined name 'config'
EMAIL_PASSWORD = config('INBOX_PASSWORD')
                 ^
./research/src/fetch_sex_places.py:94:43: F821 undefined name 'config'
        self.url = GooglePlacesURL(key or config('GOOGLE_API_KEY'))
                                          ^
./research/src/fetch_sex_places.py:253:57: F821 undefined name 'url'
            logging.info('Timeout raised for {}'.format(url))
                                                        ^
3     F821 undefined name 'config'
```

cuducos:
Hi @cclauss,

If you follow [the instructions](https://github.com/okfn-brasil/serenata-de-amor#settings) this errors go away. In other words, copy `contrib/.env.sample` as `.env`, and configure it properly before running the project ; )

Feel free to re-open the issue if you spot further problems, ok?

PR #339, opened 2018-03-07
Declare global variables that are created elsewhere

cclauss:
Declare __config__ and __url__ as global variables because they are created within the [installation process](https://github.com/okfn-brasil/serenata-de-amor#installing).  The global declaration and associated comment helps both humans and linters make sense of this non-intuitive situation.    See #338

### DIFF
--- a/research/src/fetch_sex_places.py
+++ b/research/src/fetch_sex_places.py
@@ -91,6 +91,7 @@ class SexPlacesNearBy:
         :param company: (dict) Company with name, cnpj, latitude and longitude
         :param key: (str) Google Places API key
         """
+        global config  # config is define in the installation process
         self.url = GooglePlacesURL(key or config('GOOGLE_API_KEY'))
 
         self.company = company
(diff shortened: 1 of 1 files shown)


research/src/fetch_sex_places.py
```diff
@@ -91,6 +91,7 @@ def __init__(self, company, key=None):
         :param company: (dict) Company with name, cnpj, latitude and longitude
         :param key: (str) Google Places API key
         """
+        global config  # config is define in the installation process
```
cuducos: `config` is not a global variable. It should be imported: `from decouple import config`.

research/src/fetch_sex_places.py
```diff
@@ -242,6 +243,7 @@ def parse(self, keyword, content):
         :param place: dictionary with id key.
         :return: dictionary updated with name, address and phone.
         """
+        global url  # url is define in the installation process
```
cuducos: We don't need global `url` here. It's only used for logging and there (line 255) the proper value would be `self.url.details(place_id)`. Maybe define in this line (246) `url = self.url.details(place_id)` and then user the `url` variable in lines 253 and 255.

cuducos:
The issues target by this PR are fixed in #348 (merged).

PR #341, opened 2018-03-08
Fix Rosie instalation via Docker Compose

cuducos:
Not sure why we added the directory `rosie/` to the `setup` script. I
tried it today and it didn't work – removing it, it works. Can anyone else test it please?

### DIFF
--- a/rosie/setup
+++ b/rosie/setup
@@ -4,6 +4,6 @@ import pip
 from shutil import copyfile
 
 
-copyfile('rosie/config.ini.example', 'rosie/config.ini')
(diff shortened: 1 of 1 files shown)


cuducos:
It looks like that without the `rosie/` in the `setup` script it works on the Docker flow, but not in Travis…

Irio:
Tests passing locally. Maybe aren't passing in the CI because they are run from outside Docker?

cuducos:
Yep… should we move to a dockerized CI infra?

Irio:
I think so.

cuducos:
Closed in favor of #343

PR #343, opened 2018-03-10
Dockerfiles review

Irio:
This branch has two intentions:

1. Move required environment variables to the `.env` file.
    In the master branch, environment variables are defined in the `.env`, `.env.prod`, `docker-compose.yml`, `docker-compose.production.yml`, `docker-compose.override.yml`, `.travis.yml` and when the app is already loading, as in `jarbas/settings.py`. The unification comes to follow the [DRY principle](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself).
1. Move everything production-specific to `docker-compose.production.yml` and development-specific to `docker-compose.override.yml`. As far as I understand, this pattern was already being followed, so just did a review for enforcing it.

Since I added environment variables as `ARG`s in Dockerfiles, still need to check if they can be configured in Docker Hub registry for auto building the images from master. (it works locally)

Ideally, this is a work on the direction of getting the project easier to run and, by consequence, easier to configure automated deployment from the CI.


contrib/.env.sample
```diff
@@ -1,18 +1,26 @@
-ALLOWED_HOSTS=localhost,127.0.0.1,0.0.0.0
-DEBUG=False
-SECRET_KEY=jaa-aaa-ar-bas
+ALLOWED_HOSTS=*
+DEBUG=True
```
cuducos: I tend to prefer `DEBUG=False` as default for security reasons: if someone accidentally deploy with the default settings, there's no risk of exposing too much about the system in a production environment.

docker-compose.yml
```diff
@@ -4,29 +4,19 @@ services:
...
-      - "./contrib/data:/mnt/data"
+    environment:
+      - NEW_RELIC_APP_NAME=Jarbas (Django); Jarbas (Combined)
     hostname: django
 
   tasks:
     image: datasciencebr/jarbas-backend
     environment:
-      - CELERY_BROKER_URL=amqp://guest:guest@queue/
       - NEW_RELIC_APP_NAME=Jarbas (Celery); Jarbas (Combined)
```
cuducos: As far as I got the idea of this PR, `NEW_RELIC_APP_NAME` should be in the env files, right?

jarbas/settings.py
```diff
@@ -23,10 +24,10 @@
 # See https://docs.djangoproject.com/en/1.10/howto/deployment/checklist/
 
 # SECURITY WARNING: keep the secret key used in production secret!
-SECRET_KEY = config('SECRET_KEY', default='my-secret')
+SECRET_KEY = config('SECRET_KEY', default=secrets.token_hex())
```
cuducos: Generating a random value here might lead the user to problems in the future (for example, if using authentication module, this might break password hashes because the secret key will change). If you need to “force” user to use a random string the best choice is to remove the `default` argument and leave it as mandatory to use whatever is in the `.env` file.

docker-compose.yml
Irio: This is a special case where the value changes from one service to another.

```
NEW_RELIC_APP_NAME=Jarbas (Django); Jarbas (Combined)
NEW_RELIC_APP_NAME=Jarbas (Celery); Jarbas (Combined)
```

cuducos: True story :cocktail:

Irio:
Already building in Docker Cloud, where I set values for bucket and region names. Just in case, I also defined default values in the Dockerfiles.

contrib/.env.sample
cuducos: > The rest of the values have defaults for a good development environment 

I got it. It's a valid claim. But my concern was related to security, not to an easy way to get things going for a development environment.

Just ponder it and I'll be fine with whatever decision you make : )

Irio:
The test suite is expected to pass after merging https://github.com/okfn-brasil/serenata-toolbox/pull/184. cc @cuducos

Irio:
I believe we're good to go here. The project is already running, as is in this branch and backed by Docker Cloud and Digital Ocean, at https://staging.jarbas.serenata.ai/.

In a following task, we can review documentation as a whole (I paid special attention only in the places mentioning what I was working on) and consider adding a [Deploy to Docker Cloud](https://docs.docker.com/docker-cloud/apps/deploy-to-cloud-btn/) button.

Irio:
Just did a full setup from scratch and main services are working with just the following:

```
$ git clone git@github.com:okfn-brasil/serenata-de-amor.git
$ cd serenata-de-amor
$ git checkout deployment
$ cp contrib/.env.sample .env
$ docker-compose up --build
```

(research service fails for reasons not touched by this pull request)

PR #349, opened 2018-03-23
Update to Django 2.0.3

anapaulagomes:
This PR solves https://github.com/okfn-brasil/jarbas/issues/296. To check which warnings would be fixed I ran `python -Wall manage.py test`.


jarbas/core/urls.py
```diff
@@ -1,11 +1,13 @@
...
 from jarbas.core.views import CompanyDetailView
 
 
+app_name = 'core'
+
 urlpatterns = [
-    url(
-        r'^company/(?P<cnpj>\d{14})/$',
+    path(
+        'company/<int:cnpj>/',
```
cuducos: I'm afraid the `int` parser here might remove the zero in CNPJ with leading-zeros… may we have `str` here (or a custom CNPJ that is a string with only digitis in it)?

anapaulagomes: Hmmm... I guess you are right. Take a look at this example:

```
A request to /articles/2005/03/ would match the third entry in the list. Django would call the function views.month_archive(request, year=2005, month=3).
```
Source: https://docs.djangoproject.com/en/2.0/topics/http/urls/#example

We have the option to keep the [regex style](https://docs.djangoproject.com/en/2.0/topics/http/urls/#using-regular-expressions) as well. What do you think @cuducos ?

cuducos: > We have the option to keep the [regex style](https://docs.djangoproject.com/en/2.0/topics/http/urls/#using-regular-expressions) as well

I think this is a very good choice for this situation. More explicit than just a general string (as in `<str:cnpj>`, and less complex than writing a custom object for that purpose!

PR #348, opened 2018-03-22
Update to toolbox v. 15

cuducos:
**What is the purpose of this Pull Request?**

One of our containers (`research` is broken, see #347) and the problem is also related to the `serenata-toolbox` package. This PR fix this container and align this repo to work with the new version of the toolbox. In othe words it depends on https://github.com/okfn-brasil/serenata-toolbox/pull/187

**What was done to achieve this purpose?**

* https://github.com/okfn-brasil/serenata-toolbox/pull/187 Fixed the toolbox side
* This PR fixes Pythonversion (changing the Docker image on `research/Dockerfile`) and the toolbox version (cleaning up `research/requirements.txt`)
* This clean up adds the new envvar `AMAZON_ENDPOINT` used by toolbox v. 15 (but totally optional)

**How to test if it really works?**

```
$ docker-compose up --build
```

PR #355, opened 2018-04-02
Unpin serenata-toolbox

cuducos:
@Irio, @anaschwendler: what do you think about unpinning `serenata-toolbox` from pyup.io so it always uses the latest version of the toolbox?


Irio:
As soon as the CI passes. :)

cuducos:
Closed in favor of #348

PR #370, opened 2018-04-26
New filter field on reimbursement view Api

ricardochaves:
Making the filter available by name of the policy makes it easy for customers to use the API and create user-friendly interfaces. For example, a chat bot where the user can follow certain person and for that he needs to do a search by name.

I placed one more option among the query string filters `congressperson_name__contains`. It is important to note that this field is already indexed in the database.
Also included was the test and updated documentation.

In commit there is an altered test file, looking at the differences it is easy to see the test, it follows the others. Run the project tests that it will run together.

I do not know who would be the best person, but technically anyone with a little experience in the project. The business part I think the owner of the repository should look at, as the name field is already indexed in the database I see no reason not to expose in the API.

### DIFF
--- a/jarbas/README.md
+++ b/jarbas/README.md
@@ -68,18 +68,18 @@ All these endpoints accepts any combination of the following parameters:
 * `order_by`: `issue_date` (default) or `probability` (both descending)
 * `in_latest_dataset` (_boolean_, `1` parses to `True`, `0` to `False`)
 * `search` (Search the value in any of the fields below)
-    * `congressperson_name`
-    * `supplier`
-    * `cnpj_cpf`
-    * `party`
-    * `state`
-    * `receipt_text`
-    * `passenger`
-    * `leg_of_the_trip`
-    * `subquota_description`
-    * `subquota_group_description`
+  * `congressperson_name`
+  * `supplier`
+  * `cnpj_cpf`
+  * `party`
+  * `state`
+  * `receipt_text`
+  * `passenger`
+  * `leg_of_the_trip`
+  * `subquota_description`
+  * `subquota_group_description`
 
-    
+
 For example:
 
 ```
--- a/jarbas/README.md
+++ b/jarbas/README.md
@@ -67,8 +67,19 @@ All these endpoints accepts any combination of the following parameters:
 * `year`
 * `order_by`: `issue_date` (default) or `probability` (both descending)
 * `in_latest_dataset` (_boolean_, `1` parses to `True`, `0` to `False`)
-* `congressperson_name__contains` (contains)
+* `search` (Search the value in any of the fields below)
+    * `congressperson_name`
+    * `supplier`
+    * `cnpj_cpf`
+    * `party`
+    * `state`
+    * `receipt_text`
+    * `passenger`
+    * `leg_of_the_trip`
+    * `subquota_description`
+    * `subquota_group_description`
 
+    
 For example:
 
 ```
@@ -83,6 +94,8 @@ This request will list:
 * sorted by the highest probability
 
 Also you can pass more than one value per field (e.g. `document_id=111111,222222`).
+
+Search field
 
 ##### `GET /api/chamber_of_deputies/reimbursement/<document_id>/same_day/`
 
--- a/jarbas/README.md
+++ b/jarbas/README.md
@@ -79,7 +79,6 @@ All these endpoints accepts any combination of the following parameters:
   * `subquota_description`
   * `subquota_group_description`
 
-
 For example:
 
 ```
@@ -94,8 +93,6 @@ This request will list:
 * sorted by the highest probability
 
 Also you can pass more than one value per field (e.g. `document_id=111111,222222`).
-
-Search field
 
 ##### `GET /api/chamber_of_deputies/reimbursement/<document_id>/same_day/`
 
--- a/jarbas/README.md
+++ b/jarbas/README.md
@@ -67,7 +67,7 @@ All these endpoints accepts any combination of the following parameters:
 * `year`
 * `order_by`: `issue_date` (default) or `probability` (both descending)
 * `in_latest_dataset` (_boolean_, `1` parses to `True`, `0` to `False`)
-
+* `congressperson_name__contains` (contains)
 
 For example:
 
--- a/jarbas/chamber_of_deputies/querysets.py
+++ b/jarbas/chamber_of_deputies/querysets.py
@@ -3,8 +3,9 @@ from functools import reduce
 
 from django.db import models
 from django.db.models import Q
-
+from django.db.models import F
 from django.contrib.postgres.search import SearchQuery
+from django.contrib.postgres.search import SearchRank
 
 
 class ReimbursementQuerySet(models.QuerySet):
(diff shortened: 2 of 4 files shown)


cuducos:
Hi @ricardochaves, many thanks for the interest and contribution : )

I do understand the need to improve the API and being able to lookup by the name of the congressperson is indeed a great feature.

However I think using `LIKE` queries (what `__contains` implements under the hood) is quite heavy on the database – specially when we're already optimizing searches like that with a seach vector ([used in the dashboard](https://github.com/okfn-brasil/serenata-de-amor/blob/master/jarbas/dashboard/admin.py#L374-L379)).

So what about refactoring this feature to use [the seach vector field instead](https://github.com/okfn-brasil/serenata-de-amor/blob/master/jarbas/chamber_of_deputies/models.py#L82)? It would cover the lookup in [names and more](https://github.com/okfn-brasil/serenata-de-amor/blob/master/jarbas/chamber_of_deputies/management/commands/searchvector.py#L15-L24).

jarbas/README.md
```diff
@@ -67,7 +67,7 @@ All these endpoints accepts any combination of the following parameters:
 * `year`
 * `order_by`: `issue_date` (default) or `probability` (both descending)
 * `in_latest_dataset` (_boolean_, `1` parses to `True`, `0` to `False`)
-
+* `congressperson_name__contains` (contains)
```
cuducos: This became outdated, but probably (I mean, if you agree with my feedbacl) it wil become a generic `serch` field. It would be awesome if we document all fields included in this search ; )

ricardochaves:
I made another commit before reading your answer, I'll take it easy and check `` `queryset.py```.
As soon as I have more, I'll make another commit.

ricardochaves:
@cuducos , tudo bem?

I found the [get](https://github.com/okfn-brasil/serenata-de-amor/blob/master/jarbas/chamber_of_deputies/views.py#L16
) with a lot of complexity before even adding more complexity rsrs.

CodeClimate is already accusing this, I think it would be a good time to do a refactor in it, as I have little experience in the project I will suggest here a solution, if you like I implement in this PR.

I thought of creating a ```get_filters``` that would return ```filters```. This will remove some of the complexity there.

What do you think of this approach?

About ```QuerySearch```, it took me a while to realize I needed to run the ```searchvector``` command before doing the search, but now it's the same as the dashboard

jarbas/README.md
```diff
@@ -84,6 +95,8 @@ This request will list:
 
 Also you can pass more than one value per field (e.g. `document_id=111111,222222`).
 
+Search field
+
```
cuducos: Are those two lines needed? I felt they ended up kind of meaningless. Can you check this please?

jarbas/chamber_of_deputies/querysets.py
```diff
@@ -49,6 +51,10 @@ def tuple_filter(self, **kwargs):
     def was_ordered(self):
         return bool(self.query.order_by)
 
+    def search_vector(self, search_term):
+        query = SearchQuery(search_term, config='portuguese')
+        return self.filter(search_vector=query)
```
cuducos: In the name of DRY I would include [the rank](https://github.com/ricardochaves/serenata-de-amor/blob/eb22d465b487060c2c10b971397a08a58b9aa1f6/jarbas/dashboard/admin.py#L374-L376) here and use this queryset also in the `admin.py`. If you think this is too much for this PR, would you open a new PR so we can refactor this later?

jarbas/chamber_of_deputies/tests/test_reimbursement_view.py
```diff
@@ -1,8 +1,13 @@
+import sys
+
```
cuducos: I see no need for this blank line separating `sys` from other native Python module imports. What do you think about it?

jarbas/chamber_of_deputies/tests/test_reimbursement_view.py
```diff
@@ -1,8 +1,13 @@
...
 from json import loads
 from unittest.mock import patch
 from urllib.parse import urlencode
 
+from django.core.management import call_command
 from django.shortcuts import resolve_url
+from django.utils.six import StringIO
+
```
cuducos: I see no need for this blank line separating one block of third-party imports from other third-party imports. What do you think about it?

jarbas/chamber_of_deputies/tests/test_reimbursement_view.py
```diff
@@ -22,6 +27,13 @@ def get_reimbursement(**kwargs):
     return mixer.cycle(quantity).blend(Reimbursement, search_vector=None, **kwargs)
 
 
+def create_search_vector():
+    # Running the command silently
```
cuducos: What about using a proper [docstring](https://www.python.org/dev/peps/pep-0257/) instead of a comment?

```python
def create_search_vector():
    """Running the command silently"""
    …
```

jarbas/chamber_of_deputies/tests/test_reimbursement_view.py
```diff
@@ -116,12 +128,31 @@ def test_content_with_year_filter(self):
         content = loads(resp.content.decode('utf-8'))
         self.assertEqual(3, len(content['results']))
 
-    def test_content_with_congressperson_name_filter(self):
+    def test_content_with_search_vector_one_field_filter(self):
         get_reimbursement(congressperson_name='FULANO SILVA BELTRANO', applicant_id=221)
         get_reimbursement(congressperson_name='SILVA TESTA NOME', applicant_id=23)
         get_reimbursement(congressperson_name='BELTRANO CHAVES', applicant_id=23)
+
+        create_search_vector()
```
cuducos: It's ok to use the auxiliar function you've created, totally OK. But just in case I think such call wouldn't be troublesome:

```python
call_command('searchvector', stdout=StringIO())
```

In case you considering simplofying the code and using `call_command` straightforward, just be consistent and change line 152 too : ) But, again, just if you like this idea better. The auxiliar function has its pros too : )

jarbas/chamber_of_deputies/views.py
```diff
@@ -77,25 +44,37 @@ def _build_filters(self):
         if in_latest is not None:
             self.queryset = self.queryset.in_latest_dataset(in_latest)
 
+        # filter search_vector
+        vector_param = self._vector_param()
```
cuducos: This `self._vector_param()` could be replaced by `self.request.query_params.get('search')` (and we wouldn't need the auxiliar method). It's up to you : )

jarbas/chamber_of_deputies/views.py
```diff
@@ -77,25 +44,37 @@ def _build_filters(self):
...
             self.queryset = self.queryset.tuple_filter(**filters)
 
-    def _build_vector_filters(self):
-        """ 
-            Builder and aplly all vector filters on QuerySet 
-        """
+        # change ordering if needed
+        order_by = self.request.query_params.get('order_by')
+        if order_by == 'probability':
+            self.queryset = self.queryset.order_by_probability()
```
cuducos: We might want to sort by `rank` if:

* `search` parameter is in use
* We implemented it in the `search_vector` queryset
*  User's passed no `order_by`

What do you think?

jarbas/chamber_of_deputies/views.py
```diff
@@ -77,25 +44,37 @@ def _build_filters(self):
...
-            self.queryset = self.queryset.annotate(search=SearchVector(vector)).filter(search=vector_filters[vector])
+    def _bool_param(self, param):
+        if param not in self.request.query_params:
+            return None
+
+        value = self.request.query_params[param]
+        if value.lower() in ('1', 'true'):
+            return True
+
+        return False
```
cuducos: We could try to delete lines 64-66 by replacing `value = self.request.query_params[param]` by `value = self.request.query_params.get(param)` I guess. This method could look like:

```python
    def _bool_param(self, param):
        value = self.request.query_params.get(param)
        return value.lower() in ('1', 'true')
```

cuducos:
Neat :tada: 

Many thanks for the rounds of code review – and mainly for the nice new feature : )

PR #372, opened 2018-04-29
Fix Jarbas docs on CONTRIBUTING.md

cuducos:
Following @rogeriochaves's feedback on https://github.com/okfn-brasil/serenata-de-amor/pull/365#issuecomment-385259348


anaschwendler:
I've never tried this, but it is working :)

rogeriochaves:
can confirm that it is working ;)

PR #389, opened 2018-06-01
Bugfix and enhancements on coercing dates stage

viniciusartur:
Big picture: to improve the stage of coercing dates.
- Specify a format when converting issue_field do datetime, to make clear what format we expect as input.
- Fix a bug on coercing situation_date and tuning performance of this stage.
 The bug is after coercing situation_date without specifying the format, many dates were with day/month instead month/day, (ex: 05/11/2017 were coercing to 2017-05-11).
  The initial focus of my work was tuning the performance of the coerce situation_date stage:
 - a DataFrame with 1860489 items was taking 140 seconds to coerce situation_date. After specifying the format, it was taking 10 seconds

  It was specified a format when coercing these date fields

  There are new unit tests. Just run pytest.
  You can also run rosie.py run chamber_of_deputies to check the performance of this stage and final output is the same (suspicions.xz)

  @cuducos @Irio

### DIFF
--- a/rosie/rosie/chamber_of_deputies/adapter.py
+++ b/rosie/rosie/chamber_of_deputies/adapter.py
@@ -112,6 +112,6 @@ class Adapter:
         df['is_party_expense'] = df['congressperson_id'].isnull()
 
     def coerce_dates(self, df):
-        for field in ('issue_date', 'situation_date'):
+        for field, fmt in (('issue_date', None), ('situation_date', '%d/%m/%Y')):
             self.log.info(f'Coercing `{field}` fields to date data type')
-            df[field] = pd.to_datetime(df[field], errors='coerce')
+            df[field] = pd.to_datetime(df[field], format=fmt, errors='coerce')
--- a/rosie/rosie/chamber_of_deputies/adapter.py
+++ b/rosie/rosie/chamber_of_deputies/adapter.py
@@ -112,6 +112,6 @@ class Adapter:
         df['is_party_expense'] = df['congressperson_id'].isnull()
 
     def coerce_dates(self, df):
-        for field, fmt in (('issue_date', None), ('situation_date', '%d/%m/%Y')):
+        for field, fmt in (('issue_date', '%Y-%m-%d'), ('situation_date', '%d/%m/%Y')):
             self.log.info(f'Coercing `{field}` fields to date data type')
             df[field] = pd.to_datetime(df[field], format=fmt, errors='coerce')
--- a/rosie/rosie/chamber_of_deputies/tests/test_adapter.py
+++ b/rosie/rosie/chamber_of_deputies/tests/test_adapter.py
@@ -8,6 +8,7 @@ import pandas as pd
 from freezegun import freeze_time
 
 from rosie.chamber_of_deputies.adapter import Adapter
+from datetime import date
 
 
 FIXTURES = Path() / 'rosie' / 'chamber_of_deputies' / 'tests' / 'fixtures'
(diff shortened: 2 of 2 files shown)


rosie/rosie/chamber_of_deputies/adapter.py
```diff
@@ -112,6 +112,6 @@ def rename_categories(self, df):
         df['is_party_expense'] = df['congressperson_id'].isnull()
 
     def coerce_dates(self, df):
-        for field in ('issue_date', 'situation_date'):
+        for field, fmt in (('issue_date', None), ('situation_date', '%d/%m/%Y')):
```
cuducos: Is there a specific reason for this `None` here? I think _explicit is better than implicit_ so if we're going with the default, may we have the default value instead of just `None`?

rosie/rosie/chamber_of_deputies/tests/test_adapter.py
```diff
@@ -82,3 +82,11 @@ def test_update(self, reimbursements, fetch):
             call()()
         ))
         fetch.assert_called_once_with(Adapter.COMPANIES_DATASET, self.temp_path)
+
+    @freeze_time('2010-11-12')
```
cuducos: Do we need to freeze time here? If we're reading from fixtures, _maybe_ we don't…

rosie/rosie/chamber_of_deputies/tests/test_adapter.py
```diff
@@ -82,3 +82,11 @@ def test_update(self, reimbursements, fetch):
...
+    @patch('rosie.chamber_of_deputies.adapter.fetch')
+    @patch('rosie.chamber_of_deputies.adapter.Reimbursements')
+    def test_coerce_situation_date(self, reimbursements, fetch):
+        adapter = Adapter(self.temp_path)
+        df = adapter.dataset
+        self.assertTrue(any(dt.month == 9 and dt.day == 6 for dt in df['situation_date']))
```
cuducos: In find this assertion kind of confusing… does it says that at leas one of the dates in `df['situation_date']` should be YYYY-09-06? Why are just testing the day and month and not the year? I think something like that would be more expressive for example:

```python
self.assertIn(date(2012, 9, 6), df.situation_date)
```

rosie/rosie/chamber_of_deputies/adapter.py
viniciusartur: How I can do this?

rosie/rosie/chamber_of_deputies/tests/test_adapter.py
viniciusartur: I'll test without the freeze time so. Thanks

rosie/rosie/chamber_of_deputies/tests/test_adapter.py
viniciusartur: Thanks for a outsider view of my code :)

rosie/rosie/chamber_of_deputies/adapter.py
viniciusartur: Well, As we discussed internally, I will expand this PR to also consider issue_date.

rosie/rosie/chamber_of_deputies/tests/test_adapter.py
```diff
@@ -8,6 +8,7 @@
 from freezegun import freeze_time
 
 from rosie.chamber_of_deputies.adapter import Adapter
+from datetime import date
```
cuducos: Can we keep this import section according to PEP8?

First native modules, than third-party, and lastly our own modules. That is to say:


```python
import shutil
from datetime import date
from pathlib import Path
from tempfile import mkdtemp
from unittest import TestCase
from unittest.mock import call, patch

import pandas as pd
from freezegun import freeze_time 

from rosie.chamber_of_deputies.adapter import Adapter
```

; )

viniciusartur: Sorry for not checking PEP8 good practices. Reviewing...

PR #401, opened 2018-07-18
fix pip.main  for new version

R3DDY97:
for this (https://github.com/okfn-brasil/serenata-de-amor/issues/400)
new version pip.main is changed to pip._internal.main  
in the new version pip 10

### DIFF
--- a/rosie/setup
+++ b/rosie/setup
@@ -1,7 +1,6 @@
 #!/usr/bin/env python3
 
 import pip
-# from shutil import copyfile
 
 pip_major_version = int(pip.__version__.split('.')[0])
 
--- a/rosie/setup
+++ b/rosie/setup
@@ -1,7 +1,7 @@
 #!/usr/bin/env python3
 
-import pip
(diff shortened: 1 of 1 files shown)


cuducos:
Hi @R3DDY97, 

Many thanks for the contribution. This will work on `pip` 10, but not on `pip` 9… so what about verifying if `pip` version is 10 (or higher) and, if not, raising a warning?

R3DDY97:
hi @cuducos 
yeah 
That's right way 
i didn't think about that  
will add pip version check conditional flow 

and 
is the copyfile import 

`from shutil import copyfile` 

required and used anywhere 
or can we remove it

rosie/setup
```diff
@@ -1,7 +1,12 @@
 #!/usr/bin/env python3
 
 import pip
-from shutil import copyfile
+# from shutil import copyfile
```
cuducos: I think we can delete this line (not only comment it) 👍

PR #412, opened 2018-09-04
Update interface with Rosie

Irio:
This pull request intents do update Jarbas' interface with reimbursements data to the most recent version exported by serenata-toolbox and rosie.

Since the end goal was to ensure everything works under a testing environment before pushing to production, I also made changes to the Docker setup. The current production environment was created and is still managed through Docker Cloud, discontinued in May. Therefore, this staging environment could ultimately become the new production.

The staging website is available at https://staging.jarbas.serenata.ai. At this moment, the server is running the task for loading suspicions into the database (reimbursements data should already be there).


Irio:
The serenata/elm Docker image [was not building](https://hub.docker.com/r/serenata/elm/builds/) for a month. Updating the dependencies (and removing a nuked package version) fixed the issue.

jarbas/settings.py
```diff
@@ -27,8 +27,8 @@
 # SECURITY WARNING: keep the secret key used in production secret!
 SECRET_KEY = config('SECRET_KEY')
 
-# SECURITY WARNING: don't run with debug turned on in production!
-DEBUG = config('DEBUG', default=True, cast=bool)
```
cuducos: The [`DEBUG` setting](https://docs.djangoproject.com/en/2.1/ref/settings/#std:setting-DEBUG) is a convention in Django and many third-parties depends on it (WhiteNoise, Debug Toolbar etc). I would not recommend to delete, but we could safely remove it from the environment and use the `ENVIRONMENT` to deduce if debug mode is on or off:

```python
DEBUG = config('ENVIRONMENT', cast=lambda x: x == 'development')
```

What do you think?

CONTRIBUTING.md
```diff
@@ -35,7 +35,7 @@ This is an example to run only Jarbas. First run migrations and provision:
 
 ```console
 $ docker-compose run --rm django python manage.py migrate
-$ docker-compose run --rm django python manage.py reimbursements /mnt/data/reimbursements_sample.xz
+$ docker-compose run --rm django python manage.py reimbursements /mnt/data/reimbursements-2018.csv
```
cuducos: The `reimbursement_sample.xz` is committed as a way to make it easy to get a functional version of Jarbas. Are we committing a new sample named  `reimbursements-2018.csv`? I don't see it in this PR so I'm not sure what we're attempting here.

contrib/.env.sample
```diff
@@ -1,9 +1,12 @@
...
-DEBUG=True
+CACHE_BACKEND=django.core.cache.backends.memcached.MemcachedCache
+ENVIRONMENT=development
 HTTPS_METHOD=redirect
+LOG_LEVEL=debug
 SECRET_KEY=my-secret
 USE_X_FORWARDED_HOST=False
-CACHE_BACKEND=django.core.cache.backends.memcached.MemcachedCache
+WEB_TIMEOUT=60
+WEB_WORKERS=2
```
cuducos: Can we add these new variables (and remove deprecated ones) to [Jarba's settings](https://github.com/okfn-brasil/serenata-de-amor/tree/master/jarbas#settings) docs?

jarbas/settings.py
```diff
@@ -69,7 +69,7 @@
     'debug_toolbar.middleware.DebugToolbarMiddleware',
 ]
 
-if not DEBUG:
+if ENVIRONMENT == 'production':
```
cuducos: Actually we don't need that. If I'm not wring, WhiteNoise auto-disables itself if `DEBUG == True`, so we need to simply re-add the `DEBUG` setting.

jarbas/urls.py
```diff
@@ -30,7 +30,7 @@
              namespace='chamber_of_deputies'))
 ]
 
-if settings.DEBUG:
+if settings.LOG_LEVEL == 'debug':
```
cuducos: Again, we need the `settings.DEBUG`, so let's keep it how it was ; )

jarbas/chamber_of_deputies/fields.py
```diff
@@ -1,6 +1,19 @@
...
+
 from rows import fields
 
 
+class FloatField(fields.FloatField):
+
+    @classmethod
+    def deserialize(cls, value, *args, **kwargs):
+        try:  # Rows cannot convert values such as '14,96' to float
+            value = float(value.replace(',', '.'))
```
vitorbaptista: What if `value == 1.234,56`?

cuducos: This is a valid claim indeed. But as we kind of control the input data (via `serenata-toolbox`) we can assure there's no thousands separator. But this would be a good issue for first time contributors IMHO ; )

jarbas/settings.py
Irio: Since I've been using the `ENVIRONMENT` for using or not the WhiteNoise middleware, I believe this will address your concerns in the same way.

```diff
diff --git a/jarbas/settings.py b/jarbas/settings.py
index 931e9949..802ec795 100644
--- a/jarbas/settings.py
+++ b/jarbas/settings.py
@@ -29,6 +29,7 @@ SECRET_KEY = config('SECRET_KEY')

 ENVIRONMENT = config('ENVIRONMENT', default='development')
 LOG_LEVEL = config('LOG_LEVEL', default='debug')
+DEBUG = ENVIRONMENT != 'production'

 ALLOWED_HOSTS = config('ALLOWED_HOSTS', default='*', cast=Csv())
 INTERNAL_IPS = ('127.0.0.1',)
```

jarbas/settings.py
Irio: Are you sure about this? At second sight, the middleware does not fully disable itself for this reason. Sources [[1]](https://github.com/evansd/whitenoise/blob/master/whitenoise/middleware.py) and [[2]](https://github.com/evansd/whitenoise/blob/master/docs/django.rst).

jarbas/chamber_of_deputies/fields.py
Irio: At least at this moment of time/git tree, this case won't be triggered. The file is generated by the default Pandas csv exporter, which does not add a thounsands separator. [/serenata_toolbox/chamber_of_deputies/reimbursements_cleaner.py#L149](https://github.com/okfn-brasil/serenata-toolbox/blob/57c85dae6c5c825059ccd7cbee397f1fa4e45bba/serenata_toolbox/chamber_of_deputies/reimbursements_cleaner.py#L149)

PR #428, opened 2018-10-26
Fixes cryptiles and hoek versions on package.json

jtemporal:
Fixes the security alerts on github

 

Fix the node vulnerability alerts on github

Added the suggested versions on `packaged.json` to the depencies list `"cryptiles": ">=4.1.2", "hoek": ">=4.2.1"`

```
npm install
npm run assets
```

@cuducos 

**TODO**

- check if the alert stops showing on GitHub (admin privileges required)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -7,7 +7,7 @@
       "resolved": "https://registry.npmjs.org/ansi-colors/-/ansi-colors-1.1.0.tgz",
       "integrity": "sha512-SFKX67auSNoVR38N3L+nvsPjOE0bybKTYbkf5tRvushrAPQ9V75huw0ZxBkKVeRU9kqH3d6HA4xTckbwZ4ixmA==",
       "requires": {
-        "ansi-wrap": "0.1.0"
+        "ansi-wrap": "^0.1.0"
       }
     },
     "ansi-gray": {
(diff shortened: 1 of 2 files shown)


cuducos:
> Added the suggested versions on packaged.json to the depencies list `"cryptiles": ">=4.1.2", "hoek": ">=4.2.1"`

Hi @jtemporal, many thanks for that. Unfortunately I'm not sure if this is the best alternative. Adding these packages to our `package.json` semantically says we directly depend on these packages, which is not the case. In an ideal world our dependencies would have updated their dependencies to use more secure versions of these packages, but unfortunately it looks like that's not the case.

In my opinion the best option would be to use @rogeriochaves's [`npm-force-resolutions`](https://github.com/rogeriochaves/npm-force-resolutions) to update only the `packages-lock.json` – this way we fix the vulnerability without leading others to believe we have `cryptiles` and `hoek` as direct dependencies (we avoid technical debt and keep maintainability).

However I'm really terrible in good practices in the JS world 🙊. Before we reach a decision I would love to have the opinion of people fluent in the front-end world. Thus I'm tagging some devs that might shed a light on us ❤️  @anabastos @guidiego @anarute @talyssonoc (sorry to bother you peeps out of the blue, but pythonists discussing JavaScript is probably a prelude for a disaster hahaha…).

talyssonoc:
I agree with the fact that setting`cryptiles` and `hoek` directly in the dependencies may pass the wrong message to the reader and create technical debt in the long run. On the other side, I wouldn't rely on a non-default `package.json` field like `npm-force-resolutions`' `resolutions`. Is it known what  direct dependencies depend on these declared versions of `cryptiles` and `hoek`? If so, I would suggest to update these direct dependencies.

jtemporal:
I totally agree with you both on this... although npm audit and fix and upgrades all broke elm stuff... which brings me to question... 

Do we need to maintain that purple part of jarbas online? it is that part that mainly uses npm stuff, taking it out would obliterate the problem.

I believe (analytics could prove me wrong) that the dashboard is more used since it is more friendly as a search engine nowadays. What you think @cuducos ?

jtemporal:
> We can remove the Elm part and make it as a standard Django view generating the HTML though.

I agree with this :heart:

cuducos:
Fixed by aa548407 – not the ideal way, but by reasons explained before I prefer to not include non-dependencies in `package.json`.

PR #445, opened 2019-01-07
Add filter by reimbursement number (document_id)

cuducos:
After the last architecture change we are able to store and list reimbursements that were not properly reimbursed: expenses paid directly by the Chamber of Deputies and whose values are deducted from each congresspeople quota (without actually reimbursing them, since they have not properly paid for this expense).

In other words, instead of congressperson Jane Doe paying company X and claiming the reimbursement (presenting the receipt) to the Chamber of Deputies, the Chamber directly pays company X (that is the case of flight tickets and postal services, for example).

Thus, these reimbursements have no `document_id` and no receipt, so it might be useful to filter them out in some use cases.

I just added `HasReimbursementNumberFilter` to the admin.

Start the application, set a reimbursement's `document_id` to `0` (that is the standard from the 
original dataset for this kind of expense) and test the filter in the browser:

pasted_image_07_01_19_17_22

@jtemporal @giovanisleite (by the way, any better label than “ter número de reembolso”?)


jarbas/dashboard/admin/list_filters.py
```diff
@@ -63,6 +63,23 @@ def queryset(self, request, queryset):
         return filter_option.get(self.value(), queryset)
 
 
+class HasReimbursementNumberFilter(JarbasListFilter):
+
+    title = 'ter número do reembolso'
```
giovanisleite: I think that we need a better title here, maybe something like `Pago direto ao fornecedor`.

cuducos: The leading _Por_ is fixed in Django, so I changed to _Por pagamento feito direto ao fornecedor_ <3

Issue #429, opened 2018-10-26
Removing goss from dependencies


willianpaixao:
I was looking at this health check service used in the Jarbas Docker image, [goss](https://github.com/aelsabbahy/goss).

It's a small dependency (<10MB according to its author), but it can be replaced with a simple line in `jarbas/urls.py` that adds a route (for example `/healthcheck`) that simply returns 200. Additionally, it would use `curl` which is already in the base image to request the route and verify the status of the web service.

cuducos:
Just out of curiosity: if the point is to have a view that simply returns 200, can we use a regular endpoint instead?

willianpaixao:
Precisely my friend. And right now we're using a 10MB external app for that.

willianpaixao:
It could also return a json with some simple stats of the django application (eg number of active connections, etc), and this could be used for monitoring.

PR #446, opened 2019-01-07
Remove Goss (use simpler health check)

cuducos:
As described by @willianpaixao in #429, we are not making a good use of Goss, so we could replace it by something simpler and make the Docker image smaller and simpler.

As explained in #429:

* Removed Goss
* Add a `/healthcheck` endpoint
* Replaced the `HEALTHCHECK` in the `Dockerfile`

Not sure. Any ideas?

@willianpaixao @jtemporal

### DIFF
--- a/jarbas/urls.py
+++ b/jarbas/urls.py
@@ -16,9 +16,11 @@ Including another URLconf
 """
 from django.conf import settings
 from django.conf.urls import include
-from django.http import HttpResponse
 from django.shortcuts import redirect
 from django.urls import path
+
+from jarbas.core.views import healthcheck
+
 
 urlpatterns = [
     path('', lambda _: redirect(settings.HOMES_REDIRECTS_TO), name='home'),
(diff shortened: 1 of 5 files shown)


jarbas/urls.py
```diff
@@ -27,7 +28,8 @@
     path('api/chamber_of_deputies/',
          include(
              'jarbas.chamber_of_deputies.urls',
-             namespace='chamber_of_deputies'))
+             namespace='chamber_of_deputies')),
+    path('healthcheck/', lambda _: HttpResponse(), name='healthcheck'),
```
vitorbaptista: Maybe instead of only testing that the HTTP server answer requests, we could hit the DB (and possibly Celery)? I'm thinking on a simple query, like `SELECT 1;`, just to ensure the DB is responsive.

vitorbaptista:
@cuducos should the `goss.yaml` file be removed as well?

jarbas/urls.py
cuducos: Great idea for the implementation ; ) Gonna code that soon, but feel free to send a PR to this branch.

cuducos:
> @cuducos should the `goss.yaml` file be removed as well?

Ooops. My bad. Gonna get rid of it!

Issue #447, opened 2019-01-07
Automate the update process


cuducos:
**What is the problem?**
Nowadays we need to _manually_ update our dataset because we do not have money to keep a DigitalOcean droplet with 16Gb of RAM running. But we have to automate that anyway. The manual process is done roughly like that:

1. Run Rosie locally
2. Upload the two generated CSV to the production server manually with `scp`
3. Copy the two CSV to the Docker container with `docker cp`
4. Delete the files from the production server
5. Run the command to update the reimbursements
6. Run the command to update the suspicions
7. Delete the files from the container

**How can this be addressed?**

We need to warp the steps above in a script that:

1. Creates a DigitalOcean droplet with the required RAM (16Gb), they have an API for that
2. Installs Git and Docker
3. Clone this repo
4. Runs Rosie to generate the CSV
5. Run Jarbas connecting to the production database
6. Run the two update commands (reimbursements and suspicions)
7. Delete the droplet

**Who could help with this issue?**
@jtemporal @Irio @giovanisleite @willianpaixao @cabral

andrewsmedina:
I can help with this issue.

PR #449, opened 2019-01-17
Automatically run Rosie and update Jarbas

cuducos:
This work in progress PR kicks-offs the implementation for #447. More details of the implementation in the issue page.

Feel free to jump in this PR ; )

---

Gonna update this roadmap from time to time:

- [x] ~~Create a DigitalOcean droplet~~ 982a729
- [x] ~~Installs Git~~ ddbae12
- [x] ~~Install Docker~~ 431b9b04
- [x] ~~Install Docker Compose~~ 431b9b04
- [x] ~~Clone this repo~~ 44d6542 
- [x] ~~Runs Rosie to generate the CSV~~ 282c7dd & 1147ab7
- [x] ~~Run Jarbas connecting to the production database~~ 2cc72c0
- [x] ~~Run the two update commands (reimbursements and suspicions)~~ ebd05c8
- [x] ~~Delete the droplet~~ 982a729 & 0a5cc55


cuducos:
Working on this issue I realized I oversimplified the _Run the two update commands_ — the way I described before woulde nd up in data loss (namely reimbursement receipt texts, tweets and search vector). This step actually has to:

1. Backup Twitter data to re-link reimbursements later
2. Delete all data from Reimbursement model
3. Load all data from all `reimbursements-YYYY.csv` files
4. Load the `suspicions.xz` file
5. Reload receipt texts (download the CSV and import it)
6. Rebuild the search vector
7. Restores Twitter data


I'm working on a Django command to handle that using existing update commands ; )

contrib/update/Dockerfile
```diff
@@ -0,0 +1,24 @@
+FROM python:3.7.2-alpine
+
+WORKDIR /serenata-update
+
+ADD requirements.txt requirements.txt
```
CauanCabral: Just a minor note, Docker recommends the `COPY` command instead of `ADD` in situations like that:

https://docs.docker.com/develop/develop-images/dockerfile_best-practices/#add-or-copy

cuducos: Good to know ; ) Many thanks (fixed in e2b3ab05)

cuducos:
> I'm having a hard time understanding when the update command created here will be used.

@chicocvenancio, well spotted. We must call it in the `update.yml` – gonna add a commit soon. My bad.

contrib/update/update.yml
```diff
@@ -0,0 +1,110 @@
+---
+- name: create new instance
+  hosts: 127.0.0.1
```
willianpaixao: Add a hardcoded IP address is not good. Docker often changes the localhost's address, depending on the host system.
Proposed solution, either change it to localhost or have an inventory file[[1]](https://docs.ansible.com/ansible/latest/user_guide/intro_inventory.html)(latest recommended).

contrib/update/update.yml
```diff
@@ -0,0 +1,110 @@
...
+
+  tasks:
+    - name: create new ssh keys
+      digital_ocean:
+        state: present
+        command: ssh
+        name: serenata-update
+        ssh_pub_key: "{{ lookup('file', '/root/.ssh/id_rsa.pub') }}"
```
willianpaixao: Here you name an absolute path. It makes your playbook have a dependency outside of the repo.
Since it's a public key, consider adding it to the repo or get it from a server. I personally get it from Github itself.

Although, according to the documentation, I would rather initiate the droplet with an existing key (see flag `ssh_key_ids`).

contrib/update/Dockerfile
```diff
@@ -0,0 +1,24 @@
+FROM python:3.7.2-alpine
+
+WORKDIR /serenata-update
+
+COPY requirements.txt requirements.txt
+
+RUN apk update && \
+    apk add openssh-keygen openssl-dev && \
+    /usr/bin/ssh-keygen -q -t rsa -N '' -f /root/.ssh/id_rsa && \
```
willianpaixao: Oh, now I see why you hookup those files inside of the droplet.
This way you have a sort of one time SSH key running things in your droplets. Although it works, I wouldn't recommend that. You lose track of who's running what in your cloud account.

contrib/update/requirements.txt
```diff
@@ -0,0 +1,2 @@
+ansible==2.7.5
+git+https://github.com/srkiNZ84/dopy.git@master  # until dopy supports Python 3
```
willianpaixao: That's I'm saying, you can have Ansible destroying your droplet, the key doesn't have to be destroyed (if you use the same key every time). Then you can completely ditch this non-official dependency. Here you rely on `srkiNZ84` won't add any malicious code that will have access to your DO credentials.

contrib/update/Dockerfile
cuducos: True story. But that's the best I could think at the moment. Anyway, if we can sort out the SSH Key issue with another strategy (downloading from GitHub or using `ssh_key_ids`) we tackle that too ; )

contrib/update/requirements.txt
cuducos: Python 3 issue is not related to the key issue. It's related to the Docker image used in the container I installed Ansible. The point here is Python 2 end of life (scheduled for this year) so I would like to push for Python 3. Either way, we do not have to depend on a wrapper as DigitalOcean API canbe used with HTTP requests.

contrib/update/update.yml
willianpaixao: Hum, it does. As Docker Engine controls the emission of IPs, sometimes I get something slightly different than 127.0.0.1. Just remember that localhost actually is a CIDR of 8 (netmask 255.0.0.0).

contrib/update/update.yml
willianpaixao: > This sounds good. Just to be sure, you are suggesting something like curl -O https://github.com/<user.>keys, right? In this case, user could be added as an envvar, right?
Yes, but Ansible has a tag that simplifies that. I'll search through my playbooks.

willianpaixao: > Believe me, I tried (and failed). Would you like to send a PR to this PR implementing that, or pairing on that?

Hum, yeah, let me try it.

contrib/update/update.yml
cuducos: Any ideas on how to improve from `hosts: 127.0.0.1`? I mean, the Jarbas server has a fixed IP, could we use it instead, maybe set it in an envvar?

cuducos: Fixed by 1cd210110f5db236865775ae2c4714591ffaa811

contrib/update/update.yml
cuducos: Fixed by fb58719

contrib/update/requirements.txt
cuducos: Fixed by f6cb340

contrib/update/Dockerfile
cuducos: Also fixed by fb58719

Issue #448, opened 2019-01-08
Jarbas: Search by CNPJ with separators returns incomplete results


chicocvenancio:
This template is here to help you to write your issue ;)

**What is the problem?**
Searching by CNPJ returns incomplete results. 
Seeing the [latest CEAP scandal ](https://gauchazh.clicrbs.com.br/politica/noticia/2019/01/onyx-usa-notas-em-serie-de-amigo-para-receber-verba-de-gabinete-cjqn0zueo0pwl01piavlr04z5.html) we can see a [search by CNPJ with separators](https://jarbas.serenata.ai/dashboard/chamber_of_deputies/reimbursement/?q=07.654.322%2F0001-01) returns only 3 documents, while there are [80 in a search without separators](https://jarbas.serenata.ai/dashboard/chamber_of_deputies/reimbursement/?q=07654322000101).

**How can this be addressed?**
I can see a few ways to solve this:
- We can strip the separators in the search bar with JS.
- We can strip it server side.
- We can improve the search to include results of CNPJ when that pattern is queried.

**Who could help with this issue?**
I could work on this sometime next week if it isn't solved by then.

**Labels**
[bug]

cuducos:
> We can strip it server side.

I do think this is the best option: from my experience it's easier to change Django Admin's behavior with Python code than with JavaScript code – but this might be a matter of perspective.

I think what we can do is to use a regular expression matching 14 or 11 number sequences formatted as CNPJ or CPF and strip the punctuation. This could use to match those `[\d.-]{14}|[\d./-]{18}` and then replace, in the matches, with something as `re.sub('\D', '', match)` (Python code).

cuducos:
Might be helpful (playing while enjoying my coffee earlier today):


```python
In [1]: import re

In [2]: def cleanup(value):
   ...:     for document in re.findall(r"[\d.-]{14}|[\d./-]{18}", value):
   ...:         value = value.replace(document, re.sub(r"\D", "", document))
   ...:     return value
   ...:

In [3]: cleanup('Buscar por 12.345.678/0001-69 ou 123.456.789.00')
Out[3]: 'Buscar por 12345678000169 ou 12345678900'
```

PR #451, opened 2019-01-27
Support masked CNPJ/CPF on the search

anapaulagomes:

Resolves https://github.com/okfn-brasil/serenata-de-amor/issues/448. Currently is not possible to retrieve imbursements from a masked CNPJ/CPF.

Now both AdminSite and API has a method to clean masked CNPJ/CPF.

Run Jarbas and test the search by using a masked and a not masked CNPJ or CPF. Like: 07575651000159 and 07.575.651/0001-59. Both must return the same results.

@cuducos maybe? :) I've used his snipet code.

**TODO**
Searching by CNPJ/CPF is only possible if a user types the whole document (14 or 18 characters). Maybe would be nice adding support to partial CNPJ/CPF.

### DIFF
--- a/jarbas/chamber_of_deputies/serializers.py
+++ b/jarbas/chamber_of_deputies/serializers.py
@@ -1,3 +1,5 @@
+import re
+
 from rest_framework import serializers
 
 from jarbas.chamber_of_deputies.models import Reimbursement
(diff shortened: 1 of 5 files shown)


cuducos:
Beautifully fix #448 
Thanks @anapaulagomes ; )

PR #452, opened 2019-01-27
fix bug addressed when cpfcnpj has . or - when querying database

danizavtz:
https://github.com/okfn-brasil/serenata-de-amor/issues/448
Correct cases where there is some dot or dash in the cnpj_cpf property and it queryes the database

added a method filter that includes in the query a or and query the values without the dashs and dots in cnpj_cpf

There are some tests added that address that problem. It passed

@cuducos 

I think we should add a new column or a property in the model, called "cnpj_cpf_normalized" 
and add that atribute as a or to the filter, this willl solve the cases where in the database there is a values not normalized such as:
123.456.789-00 --> covered in test cases
123.456789-00
123.45678900
123456.78900
12345678900 --> covered in test cases
the cases above are covered in query when it comes from the query_params attribute, but not when the value is persisted in the database. If the value is not normalized in database it will fail.
OR we can guarantee that there is only normalized data (a.k.a. only numbers) in the backend (postgres) some ETL when adding data to the database.

### DIFF
--- a/jarbas/chamber_of_deputies/views.py
+++ b/jarbas/chamber_of_deputies/views.py
@@ -30,7 +30,6 @@ class ReimbursementListView(ListAPIView):
         values = map(self.request.query_params.get, params)
         filters = {k: v for k, v in zip(params, values) if v}
 
-
         # filter cpf or cnpj without dot and dash
         hasCfpCnpj = self._bool_param('cnpj_cpf')
         if hasCfpCnpj is not None:
--- a/jarbas/chamber_of_deputies/views.py
+++ b/jarbas/chamber_of_deputies/views.py
@@ -30,6 +30,12 @@ class ReimbursementListView(ListAPIView):
         values = map(self.request.query_params.get, params)
         filters = {k: v for k, v in zip(params, values) if v}
 
+
+        # filter cpf or cnpj without dot and dash
+        hasCfpCnpj = self._bool_param('cnpj_cpf')
+        if hasCfpCnpj is not None:
+            self.queryset = self.queryset.cnpj_cpf(request.query_params.get('cnpj_cpf'))
+
         # filter suspicions
         suspicions = self._bool_param('suspicions')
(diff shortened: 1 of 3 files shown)


jarbas/chamber_of_deputies/views.py
```diff
@@ -30,6 +30,11 @@ def get(self, request):
         values = map(self.request.query_params.get, params)
         filters = {k: v for k, v in zip(params, values) if v}
 
+        # filter cpf or cnpj without dot and dash
+        hasCfpCnpj = self._bool_param('cnpj_cpf')
+        if hasCfpCnpj is not None:
```
cuducos: Wouldn't this line be more straightforward as `is hasCpfCnpj:`?

jarbas/chamber_of_deputies/views.py
```diff
@@ -30,6 +30,11 @@ def get(self, request):
         values = map(self.request.query_params.get, params)
         filters = {k: v for k, v in zip(params, values) if v}
 
+        # filter cpf or cnpj without dot and dash
+        hasCfpCnpj = self._bool_param('cnpj_cpf')
```
cuducos: [`hasCfpCnpj` is not a recommended variable name according to PEP8](https://www.python.org/dev/peps/pep-0008/#function-and-variable-names). Also there's a typo `Cfp` instead of `Cpf` ; )

jarbas/chamber_of_deputies/views.py
```diff
@@ -30,6 +30,11 @@ def get(self, request):
...
+        if hasCfpCnpj is not None:
+            self.queryset = self.queryset.cnpj_cpf(request.query_params.get('cnpj_cpf'))
```
cuducos: You use the search fix here, but the admin site still shows the bug, as documented in the issue. You need to hook this special case to the admin site too, not only in the API.

PR #456, opened 2019-02-16
Add summary table and charts to the Dashboard

cuducos:
Based on @hakib's post (many thanks!) [How to turn Django Admin into a lightweight dashboard](https://hakibenita.com/how-to-turn-django-admin-into-a-lightweight-dashboard) I suggested an implementation for Jarbas. This might be useful for non-tech people interested in an simple overview of CEAP (cc @tatianasb and @vilapedro).

Basically followed the steps in @hakib's post.

Run Jarbas, reach `/dashboard/chamber_of_deputies/reimbursementsummary/` and expect I expect you can browse in an interface like this one:
image

Just thinking out loud depending on each one's interest and availability: @jtemporal @giovanisleite @cabral @turicas @irio @leportella

### DIFF
--- a/jarbas/dashboard/admin/__init__.py
+++ b/jarbas/dashboard/admin/__init__.py
@@ -5,8 +5,8 @@ from brazilnum.cnpj import format_cnpj
 from brazilnum.cpf import format_cpf
 from django.contrib.postgres.search import SearchQuery, SearchRank
 from django.core.cache import cache
-from django.db.models import Count, DateField, F, Max, Min, Sum
-from django.db.models.functions import Trunc
+from django.db.models import Count, F, Max, Min, Sum
+from django.db.models.functions import Concat
 from django.utils.safestring import mark_safe
 
 from jarbas.chamber_of_deputies.models import (
@@ -168,7 +168,7 @@ class ReimbursementSummaryModelAdmin(PublicAdminModelAdmin):
         list_filters.DocumentTypeListFilter,
     )
 
-    def get_next_in_date_hierarchy(self, request):
+    def get_period(self, request):
         if 'year' in request.GET:
             return 'month'
         return 'year'
@@ -204,18 +204,25 @@ class ReimbursementSummaryModelAdmin(PublicAdminModelAdmin):
             .order_by('-total_value')
         )
 
-        period = self.get_next_in_date_hierarchy(request)
-        summary_over_time = (
-            queryset
-            .annotate(period=Trunc(
-                'issue_date',
-                period,
-                output_field=DateField()
-            ))
-            .values('period')
-            .annotate(total=Sum('total_net_value'))
-            .order_by('period')
-        )  # TODO filter by year/month
+        period = self.get_period(request)
+        if period == 'year':
+            period_key = 'year'
+            summary_over_time = (
+                queryset
+                .values('year')
+                .annotate(total=Sum('total_net_value'))
+                .order_by('year')
+            )
+        else:
+            period_key = 'period'
+            summary_over_time = (
+                queryset
+                .annotate(period=Concat('year', 'month'))
+                .values('period')
+                .annotate(total=Sum('total_net_value'))
+                .order_by('year', 'month')
+            )
+
         summary_range = summary_over_time.aggregate(
             low=Min('total'),
             high=Max('total')
@@ -229,7 +236,7 @@ class ReimbursementSummaryModelAdmin(PublicAdminModelAdmin):
             'summary_total': dict(queryset.aggregate(**metrics)),
             'summary_over_time': tuple({
                 'label': period,
-                'period': row['period'],
+                'period': row[period_key],
                 'total': row['total'] or 0,
                 'percent': self.bar_chart_height(row, low, high)
             } for row in summary_over_time)
--- a/jarbas/dashboard/admin/__init__.py
+++ b/jarbas/dashboard/admin/__init__.py
@@ -1,4 +1,4 @@
-from decimal import Decimal
+from decimal import Decimal, InvalidOperation
 from hashlib import md5
 
 from brazilnum.cnpj import format_cnpj
@@ -191,7 +191,7 @@ class ReimbursementSummaryModelAdmin(PublicAdminModelAdmin):
 
         try:
             percentage = (total - low) / (high - low)
-        except ZeroDivisionError:
+        except InvalidOperation:
             percentage = Decimal('0')
 
         ratio = Decimal('1') - minimum_percentage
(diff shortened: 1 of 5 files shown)


willianpaixao:
I can only review the Python and Jinja files. It looks good to me.

cuducos:
> @cuducos do you know if it's possible to write tests to such feature?

Yes. If I'm not wrong [this talk](https://youtu.be/41ek3VNx_6Q) gives insightful ideas about how to test Admin views. However Jarbas Django Admin parts are largely untested here. Unfortunately.

jarbas/dashboard/admin/__init__.py
```diff
@@ -149,4 +155,80 @@ def get_search_results(self, request, queryset, search_term):
...
+            size = (total - low) / (high - low) * 100
+
+        corrected_size = Decimal('5') + (Decimal('0.95') * size)
+        return corrected_size
+
+    def changelist_view(self, request, extra=None):
+        response = super().changelist_view(request, extra_context=extra)
+
+        try:
+            queryset = response.context_data['cl'].queryset
```
hakib: Since I wrote the article I found that Django will execute two more unnecessary queries on the underlying model (in your case it's `Reimbursement`). If the model holds a lot of data, these queries can be eliminated, but it requires some dirty hacks. 

If you find performance to be an issue in this specific page, you can eliminate the two extra queries.

The two queries original from [within the Django admin](https://github.com/django/django/blob/master/django/contrib/admin/options.py#L1780). To eliminate them you need to fool Django into thinking the queryset is empty. But, you still want Django to pass the queryset through the `ChangeList` and apply all the filters (from `date_hierarchy` and `list_filters`).

First, you need to override the `get_queryset` function of the `ReimbursementSummaryModelAdmin` class:

```python
def get_queryset(self, request):
        return super().get_queryset(request)
```

To fool Django into thinking the queryset is empty,  you need to return a queryset that will return no results, but one that you can reverse later (after the filters were applied):

```python
def add_false_condition(qs):
    return qs.extra(where=['1=0'])

...

class ReimbursementSummaryModelAdmin(PublicAdminModelAdmin): 
   ...
   def get_queryset(self, request):
        return add_false_condition(super().get_queryset(request).using(settings.READ_REPLICA_USING))
```

This will add the false condition "1=0" to the query. Some vendors, such as PostgreSQL will not do anything when they see such condition. This is good, this is what we want.

Now, you still need the queryset with all it's filters when you fetch the data for the list view. This is where we remove the fals condition:

```python
def remove_false_condition(qs):
    clone = qs.all()
    clone.query.where.children = [
        e for e in qs.query.where.children
        if not hasattr(e, 'sqls') or e.sqls != ['1=0']
    ]
    return clone

...

class ReimbursementSummaryModelAdmin(PublicAdminModelAdmin): 
   ...
   def changelist_view(self, request, extra_context=None):
       response = super().changelist_view(request, extra_context=extra_context)

      try:
          qs = response.context_data['cl'].queryset
      except (AttributeError, KeyError):
          return response

       qs = remove_false_condition(qs)
```

Feel free to use this but be aware that this is a really dirty hack. I only had to use this hack after our app got *really* big.

jarbas/dashboard/templatetags/dashboard.py
```diff
@@ -11,3 +13,47 @@ def rename_title(title):
...
+def brazilian_reais(value):
+    return f'R$ {brazilian_float(value)}'
+
+
+@register.filter()
+def brazilian_float(value):
+    # TODO properly use locales
+    value = value or 0
+    value = f'{value:,.2f}'
+    return value.replace(',', 'x').replace('.', ',').replace('x', '.')
```
hakib: Found this function interesting so thought I see what it does. 

Another option is using `translate`:

```
>>> brazilian_float_trans = str.maketrans(',.', '.,')                      

>>> '{:,.2f}'.format(10000 / 3).translate(brazilian_float_trans)           
'3.333,33'

>>> '{:,.2f}'.format(10000 / 3)                                            
'3,333.33'
```

jarbas/dashboard/admin/__init__.py
```diff
@@ -149,4 +155,80 @@ def get_search_results(self, request, queryset, search_term):
...
+            .values('period')
+            .annotate(total=Sum('total_net_value'))
+            .order_by('period')
+        )
+        summary_range = summary_over_time.aggregate(
+            low=Min('total'),
+            high=Max('total')
+        )
+        high = summary_range.get('high', 0)
+        low = summary_range.get('low', 0)
```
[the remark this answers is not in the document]
cuducos: Hum… interesting. After all scaling the application is easier (and cheaper) than scaling the database 👍  Done in ab46a26

jarbas/dashboard/admin/__init__.py
cuducos: We do have a bottleneck in Postgres, so we use a some caching strategies. For this summary I'm caching the whole context we calculate in b0c4ab3. That should be enough, but gonna monitor performance once it rolls out.

jarbas/dashboard/templatetags/dashboard.py
cuducos: Wow! That's way better! Refactored in 0eb34fe

cuducos:
Many thanks @willianpaixao and @hakib for the feedback. I believe addressed most of the issues and now we have a better _pull request_, yay ❤️

I did a quite relevant change dismissing `data_hierarchy`. As explained in 7457961 :

> The `date_hirearchy` method leads to doubts as `issue_date` is highly imprecise (there are expenses from 1959 entil 2050, or something like that). Also this might lead to confusion when reading the summary.

Now the UI looks like that and users can browser using regular list filters from the standard dashboard:

image

giovanisleite:
Whooa, this is so great 🎆

jarbas/dashboard/admin/__init__.py
```diff
@@ -149,4 +157,120 @@ def get_search_results(self, request, queryset, search_term):
...
+        list_filters.SuspiciousListFilter,
+        list_filters.HasReimbursementNumberFilter,
+        list_filters.StateListFilter,
+        list_filters.YearListFilter,
+        list_filters.MonthListFilter,
+        list_filters.DocumentTypeListFilter,
+    )
+
+    def get_period(self, request):
+        if 'year' in request.GET:
```
giovanisleite: This is confuse but I don't have a better suggestion 🦀
Maybe replace the `get_period` by something less context dependent. What do you think?

cuducos: Actually I think I did the opposite: I changed the name for something more context specific. But I added nice docstrings to help the reader… what do you think? It's basically c1ae375

giovanisleite: This is a better idea, very nice, @cuducos 🚀

PR #473, opened 2019-05-25
Add step to run django migrations in setup

danielSbastos:
Hi,
Here we simply add a step in `CONTRIBUTING.md` to run the project migrations before spinning up the containers.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -37,7 +37,6 @@ $ docker-compose run --rm rosie python rosie.py run chamber_of_deputies
 This is an example to run only Jarbas. First run migrations and provision:
 
 ```console
-$ docker-compose run --rm django python manage.py migrate
 $ docker-compose run --rm django python manage.py reimbursements /mnt/data/reimbursements-2018.csv
 $ docker-compose run --rm django python manage.py companies /mnt/data/companies_sample.xz
 $ docker-compose run --rm django python manage.py suspicions /mnt/data/suspicions_sample.xz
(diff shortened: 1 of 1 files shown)


cuducos:
Many thanks, but this kind of duplicate this instruction as it is already there [a few lines ahead](https://github.com/danielSbastos/serenata-de-amor/blob/d7c7f2f9fd193254f885c2c599b3a31bb11ea50d/CONTRIBUTING.md#running-jarbas). Any ideas to avoid duplication?

danielSbastos:
Oh, sorry, you're totally right, I must've passed that line when I following this guide :/

Actually, thinking more about this now, what do you think of adding the migrate command in `docker-compose.yml` to be executed before the server is spinned up? With this, we could remove it from here and the user wouldn't have to deal with this.

cuducos:
> I'd prefer to leave it here, where I added, because, in my case, after the containers were up, I went to access through the web interface and got an error.

Totally agree — it makes more sense there. I wouldn't close this PR, just make sure it does not add duplicated instructions ; )

cacarrara:
Hi, guys. Perhaps we can put a note on step `Installing` mentioning the `docker-compose up` is just a health check to assure all dependencies are succesfully installed and the project is running well. Even though, to run properly `jarbas` and `rosie` there are a few more steps (migrations, for example). So we can put a link to the proper sections: `Running Rosie` and `Running Jarbas`.

danielSbastos:
@cacarrara, nice! I like you're approach and idea. Will do :)

PR #476, opened 2019-05-25
Modify rosie adapter to load files acording to starting year

lucastt:
Make it possible to to specify the starting year for the loading step of rosie algorithms. With this changes will be easier to contribute to the project with low end machines.  

Was added a function to compare the year of the CSV being loaded and the parameter STARTING_YEAR. If the year of the file is greater or equal it will be loaded, else it won't.

During the load process verify that only the files referring to years **after** the starting year or **in** the starting year were loaded.

### DIFF
--- a/rosie/rosie/chamber_of_deputies/adapter.py
+++ b/rosie/rosie/chamber_of_deputies/adapter.py
@@ -59,7 +59,8 @@ class Adapter:
         df = pd.DataFrame()
         paths = (
             str(path) for path in Path(self.path).glob('*.csv')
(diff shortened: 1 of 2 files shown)


cuducos:
I would raise a warning here. Some classifiers, such as the meal outlier, depends on the full dataset to be more accurate (it takes into account a longitudinal analysis of how much was spent in each restaurant to find the threshold for outliers). This I can only agree with that if we make sure classifiers with these feature are loading a model generated with the full dataset.

Even though I think this might add a complexity that I'm not sure would worth it, and the risk is to offer less accurate suspicions…

PR #387, opened 2018-05-30
Optimize Jarbas Dashboard using cache

cuducos:
**What is the purpose of this Pull Request?**

Jarbas Dashboard is still extremely slow and now I attempt to use cache in order to make it faster.

**What was done to achieve this purpose?**

* Cache Django Admin `COUNT` queries
* Cache `DISTINCT` queries used to build Django Admin filters such as state and year

**How to test if it really works?**

Spin up the services and browser [the dashbaord](http://0.0.0.0:8000/dashboard/chamber_of_deputies/reimbursement/) to check I haven't broken anything. If you have the Django Toolbar enabled you can check cache is working playing around with `CACHE_BACKEND` envvar.

**TODO**

We need to make sure the cache is working on production servers. I noticed it was removed from the `docker-compose.yml`, so I'm not sure if it's in our private `docker-cloud.yml`. As Docker changed its interface (there's no toggle for _Sawrm mode_ anymore) I'm still struggling to check our production stack (cc @Irio).

**Thanks**

@fcevado who helped me identify some bottlenecks 👍

PR #360, opened 2018-04-19
Optimize query in the dashboard

cuducos:
**What is the purpose of this Pull Request?**

I noticed the home of the reimbursements in the dashboard required **105** queries in order to load the reimbursement list. That was ludicrous.

**What was done to achieve this purpose?**

I used the `list_select_related` and reduced the queries from 105 to only 5. This means more speed and less overload in our database.

**How to test if it really works?**

Open the dashboard and check if all expected info is still there ; )

**Who can help reviewing it?**

@anaschwendler @Irio @jtemporal @cabral

PR #480, opened 2019-05-26
add summary of reimbursement values in reimbursements list api

artur-deluca:
This summary can help users to easily regard the total value spent with the filtered reimbursements.

I modified the ReimbursementsListAPI paginator to include the summary of values.

In this case, however, I was unsure on how to deal with the negative reimbursement values that are on the database, since those can affect the total amount. To solve this, I simply generated two additional fields with the sum of positive and negative values.

### DIFF
--- a/jarbas/chamber_of_deputies/paginators.py
+++ b/jarbas/chamber_of_deputies/paginators.py
@@ -1,0 +1,25 @@
+from rest_framework.pagination import PageNumberPagination
+from rest_framework.response import Response
+
+
+class ReimbursementListPagination(PageNumberPagination):
+    """Paginator to show the total amount of reimbursements"""
+    def get_paginated_response(self, data):
+        return Response({
+            'summary': self.get_total_net_value(data),
+            'next': self.get_next_link(),
+            'previous': self.get_previous_link(),
+            'results': data
+        })
+
+    def get_total_net_value(self, data):
+        """Calculates the sum of reimbursements"""
+        summary = {
+            'count': self.page.paginator.count,
+            'total_positive': sum([obj['total_net_value'] if obj['total_net_value'] >= 0 else 0 for obj in data]),
+            'total_negative': sum([obj['total_net_value'] if obj['total_net_value'] < 0 else 0 for obj in data])
+        }
+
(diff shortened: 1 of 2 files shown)


cuducos:
Would you mind better explaining the purpose of this pull request, @artur-deluca? Do we have a great use case in mind for showing that information?

My point is that until further discussion, I'm personally against making the Django Admin list view for reimbursements more complex. It is already one of the most expensive cycles of request/response — in spite of optimizations and cache (for example,  #360 and #387) it is Jarba's bottleneck.

artur-deluca:
My goodness, I totally forgot to add the descriptions (comment updated). My bad!

@cuducos this came as a user request to easily summarize expenditures within a query. Maybe @sergiomario can further explain this request, since he was the one that suggested to me.

I agree on the drawbacks, but are we talking about the same list view? If so, I would be glad to discuss any other alternatives!

cuducos:
> @cuducos this came as a user request to easily summarize expenditures within a query.

Maybe #456 is a better alternative for summaries…

sergiomario:
Oh sorry @artur-deluca :/

I definitely did not know the endpoint [reimbursementsummary](https://jarbas.serenata.ai/dashboard/chamber_of_deputies/reimbursementsummary/).
I think it's a usability issue. When accessing Jarbas, I believe that the user should have a way to navigate to this information, from a simple m filter.
Based on my experience, people may not know the availability of this function in Jarbas.

cuducos:
Easy way out: changing `HOMES_REDIRECTS_TO` in [`settings.py`](https://github.com/okfn-brasil/serenata-de-amor/blob/master/jarbas/settings.py#L204) will to `'/dashboard/'` (instead of current `'/dashboard/chamber_of_deputies/reimbursement/'`) will make users land at:

image

Alternatively I think a proper roadmap would be:
- [ ] implement text search in `reimbursementsummary`
- [ ] add a link to any query/search/filter in `reimbursement` to see its summarized version

artur-deluca:
What do you guys think of creating a tab structure connecting summary and table?

image

I don't know if it is totally feasible but in this case, we could simply create a template that renders both table and summary and hide the corresponding components according to the tab selected using css.

Issue #425, opened 2018-10-22
Automate searchvector creation


giovanisleite:
**What is the problem?**
Today the [searchvector](https://docs.djangoproject.com/en/2.1/ref/contrib/postgres/search/) creation is made manually with the command `python manage.py searchvector` ([code](https://github.com/okfn-brasil/serenata-de-amor/blob/be9c2cf0d2a01d8895980f4eb83b51830034a0ee/jarbas/chamber_of_deputies/management/commands/searchvector.py)) and it could be **automated**.
This command should run every **48h**, between **2 am** and **4 am**.

**How can this be addressed?**
- Use the `jarbas/celery.py` file
- Create a function that schedule the execution of the management command to be executed every 48h ([schedule function example](https://github.com/okfn-brasil/diario-oficial/blob/83a0df9dc5f7664a79b4d1faee73fc78c6a8689d/processing/tasks.py#L19) - This example schedules to every day, 13 hours, [celery documentation brings other forms of scheduling](http://docs.celeryproject.org/en/latest/userguide/periodic-tasks.html) ).
- Create the function that executes the `python manage.py searchvector` command ([execute a command example](https://github.com/okfn-brasil/diario-oficial/blob/83a0df9dc5f7664a79b4d1faee73fc78c6a8689d/processing/tasks.py#L61)).

**Who could help with this issue?**
I can help on sundays =)

**Labels**
[enhancement]

This issue is a translation of `Melhorias no Jarbas/Automatizar atualizações/1`, page 4 of [this document](https://docs.google.com/document/d/1qjYHKr9FLAaDwI4VJeHbcWj4LKrAWBeHqrB6VqBLXi8/edit#heading=h.6nco566ujoz0).

luciakasman:
Hello! I would like to work on this issue :)

giovanisleite:
Hey, @luciakasman, @oleggator already did this specific task, but in #497 there is a discussion about adding more things to the celery schedule, you could do it if you want.

Thanks for the interest!

PR #497, opened 2019-10-01
Searchvector automation

oleggator:
Resolution of issue #425.

- Added beat (celery scheduler) container to docker-compose
- Scheduled `searchvector` command execution every second day at 2am
- Add parameters to settings.py

Will be printed messages like:
```
tasks_1     | [2019-10-01 18:31:00,003: WARNING/ForkPoolWorker-2] Running searchvector...
tasks_1     | [2019-10-01 18:31:00,024: WARNING/ForkPoolWorker-2] Searchvector is done
```

@giovanisleite 

**TODO**


jarbas/celery.py
```diff
@@ -1,9 +1,34 @@
...
 os.environ.setdefault('DJANGO_SETTINGS_MODULE', 'jarbas.settings')
 
 app = Celery('jarbas')
 app.config_from_object('django.conf:settings', namespace='CELERY')
 app.autodiscover_tasks()
+
+
+@app.on_after_configure.connect
+def setup_periodic_tasks(sender, **kwargs):
+    from django.core import management
```
oleggator: `management` may be imported only after app has been initialized

sergiomario:
@giovanisleite please, can you review this? :-)

jarbas/celery.py
```diff
@@ -1,9 +1,34 @@
...
+
+    @app.task(ignore_result=True)
+    def searchvector():
+        print('Running searchvector...')
+        management.call_command('searchvector')
+        print('Searchvector is done')
```
cuducos: Can we use a propper log instead of `print`?

jarbas/celery.py
```diff
@@ -1,9 +1,34 @@
...
+
+    if settings.SCHEDULE_SEARCHVECTOR:
```
cuducos: What do you think about early abort to keep a [flat success path](https://www.vinta.com.br/blog/2018/flat-success-path/), that is to say, the main code in the main indentation level?

cuducos:
Personally I believe that this PR brings an avoidable level of complexity in terms of maintainability -- but it is definitively **not** @oleggator's fault. Basically I believe we need to adopt his design (which is better than what we have) to avoid having _two_ different ways to handle scheduled jobs in our stack.

What I mean is that, nowadays, we keep a bunch of scheduled tabs in manually set [`crontab`](https://github.com/okfn-brasil/serenata-de-amor/blob/master/contrib/crontab/crontab). As @oleggator's have a lot of advantages in terms of maintainability (basically embedding the schedule thing in the application layer, and not as a server manual configuration), I would suggest for maintainers to move all cronjobs to the design suggested in this PR ; )

jarbas/celery.py
oleggator: Replaced `print` with `logging`

jarbas/celery.py
oleggator: Reduced indentation level

oleggator:
Maybe I can modify existing cron scripts to run using celery?

PR #500, opened 2019-10-15
Remove invalid modules

tramontini:
The purpose of this PR is to clean some invalid calls from the repository. Some of the modules imports were duplicated or not being used. Because of that I only remove the invalid modules imports.

The modules removal should not interfere in the actual operation.

### DIFF
--- a/research/src/fetch_cnpj_info.py
+++ b/research/src/fetch_cnpj_info.py
@@ -13,7 +13,6 @@ import requests
 import requests.exceptions
 import re
 import logging
-import json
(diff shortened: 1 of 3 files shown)


luizfzs:
@tramontini the changes are looking good. Just as a recommendation for the future, it's a good practice to never commit directly to the master branch. So, on this case, instead of commiting directly to your fork's `master`, you would create a branch, commit the work and create a pull request from that new branch to `okfn-brasil:master`.

Issue #506, opened 2019-12-16
The meal_price_outlier_classifier classifier is broken


sergiomario:
**What is the problem?**
There is an error executing the script `pipenv run ansible-playbook update.yml`.
This script runs weekly to update the *Câmara dos Deputados* data and classify suspicious invoices. As defined in the [data update procedure](https://github.com/okfn-brasil/serenata-de-amor/tree/master/contrib/update).
![erro_rosie](https://user-images.githubusercontent.com/2698516/70950658-01e4f780-2040-11ea-83ec-a1d690fb3fca.png)

```
File serenata-de-amor/rosie/rosie/chamber_of_deputies/classifiers/meal_price_outlier_classifier.py, line 105, in __normalize_string

nfkd_form = unicodedata.normalize('NFKD', string.lower())", 
"AttributeError: 'float' object has no attribute 'lower'"
```

**How can this be addressed?**
Type validation can be done by ensuring that a String object is being manipulated.

cuducos:
In the Chamber of Deputies [adapter, we only enforce a couple of fields to be read as strings](https://github.com/okfn-brasil/serenata-de-amor/blob/55b07726a49e9ff7149b313f2c13d6199ab2f4a2/rosie/rosie/chamber_of_deputies/adapter.py#L25-L30):

```python
    DTYPE = {
        'applicant_id': np.str,
        'cnpj_cpf': np.str,
        'congressperson_id': np.str,
        'subquota_number': np.str
    }
```

Probably because in most cases Pandas would read the column `recipient` as strings as well. The error suggests it doesn't (the _traceback_ leads to a single call of the `__normalize_string` method with is called using the dataset `recipient` column).

Also, [the adapter creates the `recipient` column from the `supplier` column](https://github.com/okfn-brasil/serenata-de-amor/blob/55b07726a49e9ff7149b313f2c13d6199ab2f4a2/rosie/rosie/chamber_of_deputies/adapter.py#L23). So, my best guess to try to fix this bug would be to enforce the `np.str` as the `dtype` of the supplier, adding a single line to block I pasted above:

```python
    DTYPE = {
        'applicant_id': np.str,
        'cnpj_cpf': np.str,
        'congressperson_id': np.str,
        'subquota_number': np.str,
        'supplier': np.str,
    }
```

**However** it's extremely unlikely that Pandas is reading this column as a numeric one for no reason. We should make sure the CSV inputs still consistent to whatever we're expecting from them.

[...]

PR #507, opened 2019-12-29
Ensure company name is string in meal outlier classifier

cuducos:

To address the bug discussed in #506 

Before using a string method within the classifier, I used Pandas `fillna`  to replace `np.nan` (a _float_) by an empty string.

I added an empty value as a feature, so `python rosie.py test chamber_of_deputies` is an option. But fully running Rosie with `python rosie.py run chamber_of_deputies` is also advised.

@sergiomario @luizfzs


luizfzs:
Regarding the `safety-ci` having failed, is that check strictly mandatory?

cuducos:
> is that check strictly mandatory?

They are important, but (IMHO) in the context of #503, for example — they are not the scope og this change proposed here ; )

Anyway, @sergiomario has the final call : )

PR #533, opened 2020-08-29
add requirements-without-conda file

coelhudo:
**What is the purpose of this Pull Request?**

The purpose is to fix issue #532

**What was done to achieve this purpose?**

An alternative file, called requirements-without-conda.txt was created.

**How to test if it really works?**

Same tests described in [rosie/README.md](https://github.com/okfn-brasil/serenata-de-amor/blob/main/rosie/README.md) and [jarbas/README.md](https://github.com/okfn-brasil/serenata-de-amor/blob/main/jarbas/README.md#local-install)

Specs of the host where I tested:

- OS: Slackware current kernel 5.4.60
- Python 3.7.6
- RabbitMQ: 3.7.28
- PostgreSQL 10

**Who can help to review it?**

Anyone

PR #536, opened 2020-10-07
added requirements file for pip

abhinand5:
**Purpose of this PR**
This PR fixes issue #533 

**Changes made**
Alternative file `requirements.txt` was added.


cuducos:
I would like to share my thoughts here. There is a reason we suggest Anaconda: it is simple and it help maintainers because of its simplicity.

This `pip` version suggested here would only work if the user already has a series of dependencies installed in their OS — this would make the system ready to compile under the hood requirements for those packages listed in the new file.

This adds complexity, expand the window for frustration in first-time users, and covering that would leave instructions longer and longer. Therefore, I advice current maintiners to be careful on this choice.

abhinand5:
Fair enough.

Issue #556, opened 2021-06-29
rosie is not posting on twitter 


leao-nardo:
[twitter]
https://twitter.com/RosieDaSerenata

i just found the project , but i looked in twitter that the last post from rosie is from march 3 . is the project still running ?

or the politicians are well behaving and rosie didnt found a suspicious note since march ?

Issue #560, opened 2021-08-23
Rosie stops mid-classification due to MemoryError in a 32gb ram machine


giuliocc:
Rosie stopped tweeting a while back and that was the reason.

In the last weeks @andreformento diagnosed this locally and we tested it in the production infrastructure.

Here's the full traceback for executing `python3 rosie.py run chamber_of_deputies` in a common 8 vcpus 32gb ram Digital Ocean's Droplet:
```
2021-08-23 22:32:46,878 - rosie.chamber_of_deputies.adapter - INFO - Updating companies
Downloading 2016-09-03-companies.xz: 100%|████████████████████████████████████████████████████████████████████████████| 4.84M/4.84M [00:00<00:00, 34.5Mb/s]
2021-08-23 22:32:47,051 - rosie.chamber_of_deputies.adapter - INFO - Updating reimbursements from 2009
2021-08-23 22:33:05,802 - rosie.chamber_of_deputies.adapter - INFO - Updating reimbursements from 2010
2021-08-23 22:33:27,758 - rosie.chamber_of_deputies.adapter - INFO - Updating reimbursements from 2011
2021-08-23 22:33:52,820 - rosie.chamber_of_deputies.adapter - INFO - Updating reimbursements from 2012
2021-08-23 22:34:14,875 - rosie.chamber_of_deputies.adapter - INFO - Updating reimbursements from 2013
2021-08-23 22:34:39,627 - rosie.chamber_of_deputies.adapter - INFO - Updating reimbursements from 2014
2021-08-23 22:35:00,156 - rosie.chamber_of_deputies.adapter - INFO - Updating reimbursements from 2015
2021-08-23 22:35:24,343 - rosie.chamber_of_deputies.adapter - INFO - Updating reimbursements from 2016
2021-08-23 22:35:47,603 - rosie.chamber_of_deputies.adapter - INFO - Updating reimbursements from 2017
2021-08-23 22:36:10,159 - rosie.chamber_of_deputies.adapter - INFO - Updating reimbursements from 2018
2021-08-23 22:36:29,338 - rosie.chamber_of_deputies.adapter - INFO - Updating reimbursements from 2019
2021-08-23 22:36:47,928 - rosie.chamber_of_deputies.adapter - INFO - Updating reimbursements from 2020
2021-08-23 22:36:58,705 - rosie.chamber_of_deputies.adapter - INFO - Updating reimbursements from 2021
[...]

PR #561, opened 2021-08-23
Enlarge droplet to 64GB RAM to avoid memory error

ogecece:
When running the `/contrib/crontab/update` task, the `TraveledSpeedsClassifier` was raising `MemoryError` in it's `predict` method (line 70).

This is a temporary solve. We should investigate if this memory consumption can be reduced.

This issue is further described in #560.

Fixes:
#556

### DIFF
--- a/contrib/update/update.yml
+++ b/contrib/update/update.yml
@@ -17,7 +17,7 @@
         image_id: ubuntu-18-04-x64
         name: serenata-update
         region_id: nyc1
-        size_id: s-8vcpu-32gb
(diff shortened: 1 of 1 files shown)


cuducos:
As these droplets (virtual machines, for those not familiar with DigitalOcean jargon) are short-lived, this might not be only a temporary fix. I think it worth it to deploy and check how much the bill raises at the end of the month. If that is a considerable amount, then we could try to find a better way…

Maybe finally [it is time for Dask](https://github.com/okfn-brasil/rosie/issues/24) ✨

